PR #9, opened 2022-01-27
Avoid retrying stream payloads

frumsdotxyz:
Addresses https://github.com/deepgram/python-sdk/issues/8 by not retrying requests if the payload is a file stream that can no longer be read.


nikoladg (dismissed):
as long as it was tested, code LGTM!

geekchick (approved):

PR #22, opened 2022-04-25
Update README.md real-time streaming code snippet

geekchick:
Looks like this line `await deepgramLive.finish()` is not reachable in the current snippet. Updated code in the `while` loop to check if no data coming from live stream then break out of that loop.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201885737791795/1202176183709796) by [Unito](https://www.unito.io)


frumsdotxyz (approved):
👍

PR #23, opened 2022-05-16
Initial commit of new Python Endpoints

geekchick:
New endpoints include:

* billing
* invitations
* members
* scopes

I also made a slight change to the update method of the `projects` endpoint.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201885737791795/1202287564724308) by [Unito](https://www.unito.io)

### DIFF
--- a/deepgram/projects.py
+++ b/deepgram/projects.py
@@ -24,20 +24,16 @@ class Projects:
             headers={'Content-Type': 'application/json'}
         )
 
-    async def update(self, project_id: str, name: str = None,
-                     company: str = None) -> UpdateResponse:
+    
+    async def update(self, project_id: str, **payload) -> UpdateResponse:
         """Updates a project's information."""
-        payload = {}
-        # there's got to be a better way to do this without non-specific kwargs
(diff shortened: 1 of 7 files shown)


deepgram/projects.py
```diff
@@ -24,21 +24,17 @@ async def create(self, name: str) -> Project:
...
+    
+    async def update(self, project_id: str, **payload) -> UpdateResponse:
         """Updates a project's information."""
-        payload = {}
-        # there's got to be a better way to do this without non-specific kwargs
-        if name:
-            payload['name'] = name
-        if company:
-            payload['company'] = company
+
```
frumsdotxyz, line 30:
```suggestion
        payload = {k: v for k, v in {
            'name': name,
            'company': company,
        } if v}
        
```

I think I like the idea of staying explicit about the parameters the endpoint can take, so users don't have to refer back to the docs as much!
We can make it a little more elegant than the previous version, though :p

frumsdotxyz (approved):
Looks good! :>

PR #48, opened 2022-09-16
Update _types.py with new Topic Detection Feature

geekchick:


frumsdotxyz (approved):
This and #49 should probably also include an update to the response body type that includes the new properties, once the specifics of those are finalized.

PR #60, opened 2022-11-01
added formatting features

briancbarrow:


deepgram/_types.py
```diff
@@ -63,13 +63,20 @@ class TranscriptionOptions(TypedDict, total=False):
...
     callback: str
     keywords: List[str]
     ner: str
     tier: str
-    replace: str
+    dates: str
+    date_format: str
+    times: str
+    dictation: str
+    measurements: str
```
geekchick, line 79: Are we removing the `replace` parameter? 

Also, I wanted to double-check check but should `dates, times, dictation, measurements` all be booleans and not strings? 

Lastly, are we missing `smart_format` per the documentation?

geekchick (changes requested):
I just have a few questions about the parameter types and a few other things.

geekchick (approved):

PR #70, opened 2023-01-05
added toggle function

briancbarrow:


frumsdotxyz:
If we expect the scope of these messages to expand in the future, I wonder if it might make sense to specify more broadly as a `configure` method that can be provided an argument like `{'numerals': True}` (with a type spec to clarify what's possible) rather than a dedicated method for just numerals.

briancbarrow:
@frumsdotxyz That is a good idea. I'm still new to Python so let me know if the change I made is what you were thinking, or if I need to make more changes.

frumsdotxyz (approved):

deepgram/transcription.py
```diff
@@ -296,6 +296,13 @@ def send(self, data: Union[bytes, str]) -> None:
 
         self._queue.put_nowait((False, data))
 
+    def configure(self, config: ToggleConfigOptions) -> None:
+        """Toggles whether or not numerals are included in the transcription."""
```
frumsdotxyz, line 300:
```suggestion
        """Sends messages to configure transcription parameters mid-stream."""
```

frumsdotxyz (approved):

PR #87, opened 2023-05-17
:bug: Fix urllib error handling

### LINKED ISSUE #86: AttributeError: ‘URLError’ object has no attribute ‘status’
## What is the current behavior?

Using the following code, we get an error. This is because Deepgram Nova doesn't support Spanish. However, the error relates to parsing the response incorrectly. Which appears to be a bug.

(not the actual code from the ticket)

```python
from deepgram import Deepgram
import json
import os
​
DEEPGRAM_API_KEY = os.getenv('DEEPGRAM_API_KEY') 

PATH_TO_FILE = 'test.mp3'
MIMETYPE = 'audio/mp3'
​
def main():
    dg_client = Deepgram(DEEPGRAM_API_KEY)
    
    with open(PATH_TO_FILE, 'rb') as audio:
        source = {'buffer': audio, 'mimetype': MIMETYPE}
        options = {"utterances":True,"diarize":True,"utterances":True, "model": "nova", "language": "en-US","paragraphs":True,"smart_format":True }

// ...

        response = dg_client.transcription.sync_prerecorded(source, options)
        print(response['results'])
​
main()
```

The error:

```

// ...

AttributeError: ‘URLError’ object has no attribute ‘status’
```

jcdyer:
Fixes #86 - a defect where we catch urllib.error.URLError, and then introduce special handling for client errors based on the `status` attribute. URLError is a base class that is only guaranteed to have a `reason` attribute.  Instead we should be checking the subclass `HTTPError`, and using the `code` attribute to determine the status code.

The issue manifested as:

```python
 File “/d/code/tribal/tempo/test.py”, line 42, in <module>
  main()
 File “/d/code/tribal/tempo/test.py”, line 32, in main
  response = dg_client.transcription.sync_prerecorded(source, options)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 File “/d/code/youtube_dump/.venv/lib/python3.11/site-packages/deepgram/transcription.py”, line 355, in sync_prerecorded
  return SyncPrerecordedTranscription(
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 File “/d/code/youtube_dump/.venv/lib/python3.11/site-packages/deepgram/transcription.py”, line 111, in __call__
  return _sync_request(
      ^^^^^^^^^^^^^^
 File “/d/code/youtube_dump/.venv/lib/python3.11/site-packages/deepgram/_utils.py”, line 153, in _sync_request
  return attempt()
      ^^^^^^^^^
 File “/d/code/youtube_dump/.venv/lib/python3.11/site-packages/deepgram/_utils.py”, line 148, in attempt
  raise (Exception(f’DG: {exc}’) if exc.status < 500 else exc)
                   ^^^^^^^^^^
AttributeError: ‘URLError’ object has no attribute ‘status’
```

### DIFF
--- a/deepgram/_utils.py
+++ b/deepgram/_utils.py
@@ -144,8 +144,8 @@ def _sync_request(
                 if body.get('error'):
                     raise Exception(f'DG: {content}')
                 return body
-        except urllib.error.URLError as exc:
-            raise (Exception(f'DG: {exc}') if exc.status < 500 else exc)
+        except urllib.error.HTTPError as exc:
+            raise (Exception(f'DG: {exc}') if exc.code < 500 else exc)
 
     tries = RETRY_COUNT
     while tries > 0:


lukeocodes (approved):
Thanks Cliff, we'll get this released ASAP

PR #90, opened 2023-05-26
Fix PrerecordedOptions not being parsed correctly

UP929312:
If I'm 100% honest, I haven't particularly followed the Contributing Guidelines, so don't have an issue on the forum, but I noticed that Punctuation and the model was wrong, and after investigating, realised it was because the `_make_query_string` wasn't working as intended. After some debugging, I realised it was given a dictionary (as expected), but inside another dictionary.
That wrapper dictionary broke the query string builder, and this just unwraps it.

### DIFF
--- a/deepgram/transcription.py
+++ b/deepgram/transcription.py
@@ -56,6 +56,8 @@ class PrerecordedTranscription:
             source.get('buffer', {'url': source.get('url')})
         )
         content_type = cast(str, source.get('mimetype', 'application/json'))
+        if "option" in self.transcription_options:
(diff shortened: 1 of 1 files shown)


briancbarrow:
@UP929312 What was the code you were running when you ran into this issue? I haven't seen this issue come up before so i want to make sure this is a common issue before merging this change. 

Are you passing in a dictionary with an `options` property?

jjmaldonis:
I'm going to close this PR because a dictionary of the form `{"options": ...}` is not indented to be a valid argument to the Deepgram SDK. Please feel free to reopen if I'm missing something.

PR #124, opened 2023-08-01
fixed pytests and added a new one

jjmaldonis:
Most of these changes are formatting changes.

I added one new pytest to test transcriptions of files on disk. I uploaded the 6 second audio file to this repo.

As a side note, our tests were failing after the latest release of Nova + Smart Format because the trailing period at the end of the transcript is no longer included. I told Shir.


tests/test_transcription.py
```diff
@@ -64,3 +65,14 @@ def test_transcribe_prerecorded_find_and_replace_list():
...
     actual_transcript = response["results"]["channels"][0]["alternatives"][0]["transcript"]
     assert actual_transcript == MOCK_TRANSCRIPT.replace("fast", "slow").replace("miss", "snooze")
+
+
+@pytest.mark.asyncio
+async def test_transcribe_prerecorded_file():
+    """
+    Test basic asyncronous pre-recorded transcription.
+    """
+    example_wav_file = os.path.join(CURRENT_DIRECTORY, "the-missle-short.wav")
```
DamienDeepgram, line 75: "missile"

DamienDeepgram (changes requested):
LGTM with the spelling mistake corrected

tests/test_transcription.py
DamienDeepgram: The file "the-missle-short.wav" does not exist

tests/test_transcription.py
```diff
@@ -2,27 +2,30 @@
...
+
+CURRENT_DIRECTORY = os.path.split(os.path.abspath(__file__))[0]
+
 api_key = pytest.api_key
 assert api_key, "Pass Deepgram API key as an argument: `pytest --api-key <key> tests/`"
 
 deepgram = Deepgram(api_key)
 
-MOCK_TRANSCRIPT = "Yep. I said it before, and I'll say it again. Life moves pretty fast. You don't stop and look around once in a while. you could miss it."
+MOCK_TRANSCRIPT = "Yep. I said it before, and I'll say it again. Life moves pretty fast. You don't stop and look around once in a while. you could miss it"
```
jkroll-deepgram, line 18: I set up this test strategy - maybe there is a more forgiving way than string exact match? Like having no more than X character edit distance from expected transcript. Otherwise these small transcription changes will keep making this kind of test fail. But that could also be out of scope of this PR. What do you think?

tests/test_transcription.py
jjmaldonis: Fixed, thanks!

tests/test_transcription.py
```diff
@@ -64,3 +64,14 @@ def test_transcribe_prerecorded_find_and_replace_list():
...
     actual_transcript = response["results"]["channels"][0]["alternatives"][0]["transcript"]
     assert actual_transcript == MOCK_TRANSCRIPT.replace("fast", "slow").replace("miss", "snooze")
+
+
+@pytest.mark.asyncio
+async def test_transcribe_prerecorded_file():
+    """
+    Test basic asyncronous pre-recorded transcription.
+    """
+    example_wav_file = os.path.join(CURRENT_DIRECTORY, "the-missile-short.wav")
```
jkroll-deepgram, line 74: Can we avoid committing a 1+ MB audio file to the repo, for the sake of keeping the package lighter-weight for users (even though it's already 170 MB 😬 )? We could submit a URL of a DG-hosted file instead of the audio itself?

tests/test_transcription.py
jjmaldonis: I agree, something that allows small differences in the transcript would be nice. We can use the `fuzzywuzzy` package :) to do this. I can add it to this PR. I'll add a requirements.txt file as well since we'll have multiple dependencies for testing after this PR.

tests/test_transcription.py
jjmaldonis: Would the test then download the file to the user's computer before running the test?

jjmaldonis: We alternatively create a config file that prevents the-missle-short.wav from being included in the release package.

jkroll-deepgram: I did `AUDIO_URL = "https://static.deepgram.com/examples/Bueller-Life-moves-pretty-fast.wav"` for the other tests. Does it work to keep the pattern of just submitting the JSON to DG without having the audio locally at all?

jkroll-deepgram: Jason and I discussed separately - this test requires local audio to test the buffer, and we don't have any audio files in the repo yet. Fine as-is.

jkroll-deepgram (approved):

DamienDeepgram (approved):
LGTM

PR #126, opened 2023-08-03
Update _version.py one order

WeurnomboTroc:


jjmaldonis:
We have an internal CI/CD workflow that automatically updates this version number for new releases, but thanks for the PR!

PR #131, opened 2023-08-11
Added Tests to test for the JSON structure of the deepgram response

SaarthShah:
> Fixed the '.' causing the tests to fail
> Checking to make sure that result from the SDK is in the correct JSON format. These should not break unless there is a breaking change from the API itself.


jjmaldonis:
Hey @SaarthShah I just merged a PR with a minor conflict. Hopefully it's easy to resolve though. If you want me to review this PR let me know.

jkroll-deepgram (approved):

PR #133, opened 2023-08-11
added test cases for features & testing for JSON structure

SaarthShah:
> I added tests to make sure summarize v2, topic detection and diarization is working as expected. Also, testing the JSON structure of the overall/base response too

### DIFF
--- a/tests/test_transcription.py
+++ b/tests/test_transcription.py
@@ -152,8 +152,11 @@ def test_diarization():
         },
     )
 
-    # Checks if the speakers key is present in the alternatives[0] object
+    # Checks the keys for the alternatives[0] object
     assert set(response['results']['channels'][0]['alternatives'][0]['paragraphs']['paragraphs'][0].keys()) == set(['sentences','speaker', 'num_words', 'start', 'end'])
+
+    # Checks the keys in the words object
+    assert set(response['results']['channels'][0]['alternatives'][0]['words'][0].keys()) == set(['word', 'start', 'end', 'confidence', 'speaker', 'speaker_confidence', 'punctuated_word'])
 
 def test_summarize():
     """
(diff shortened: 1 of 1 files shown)


tests/test_transcription.py
```diff
@@ -76,3 +76,131 @@ async def test_transcribe_prerecorded_file():
...
+            "model": "nova",
+            "smart_format": True,
+            "summarize": 'v2'
+        },
+    )
+    
+    # Checks if the summary object has the expected keys
+    assert set(response['results']['summary'].keys()) == set(['result', 'short'])
+    # Check if the result is a string
+    assert type(response['results']['summary']['result']) == str
```
jjmaldonis, line 176: Is it worth it to check if the value of this is `success` rather than only if it's a string? (I think it's `success` when the request succeeds.)

jjmaldonis (dismissed):
Looks good to me! I didn't run the tests to make sure they work, so as long as they succeed after the last commit then I say go ahead and merge. I left two questions that you may want to change - at your discretion.

tests/test_transcription.py
SaarthShah: Yup, ur right. Adding a case for this!

tests/test_transcription.py
```diff
@@ -76,3 +76,131 @@ async def test_transcribe_prerecorded_file():
...
+        },
+        {
+            "model": "nova",
+            "smart_format": True,
+            "diarize": True
+        },
+    )
+
+    # Checks if the speakers key is present in the alternatives[0] object
+    assert set(response['results']['channels'][0]['alternatives'][0]['paragraphs']['paragraphs'][0].keys()) == set(['sentences','speaker', 'num_words', 'start', 'end'])
```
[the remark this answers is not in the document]
SaarthShah: It's in words but added a case for that too

jjmaldonis (dismissed):

jjmaldonis:
It looks like there is a conflict that will need to be resolved before merging, but it looks like it's a pretty simple one and the two additions to this file should both be included in the merge.

SaarthShah:
Do you know how to resolve these conflicts @jjmaldonis ??

jjmaldonis (approved):

jjmaldonis:
@SaarthShah I just resolved the changes. You can do it by clicking the "Resolve conflicts" button in GitHub, or go following these instructions on the command line / locally: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/addressing-merge-conflicts/resolving-a-merge-conflict-using-the-command-line

PR #135, opened 2023-08-16
added regex validation for API key

jjmaldonis:
This is a simple change to ensure the user has included an API key with the correct format.

Before this PR, if the API key was set incorrectly, the Python SDK threw a weird error SSL that was difficult to understand. This change makes it clear the issue is an invalid API key.

All tests pass.

### DIFF
--- a/deepgram/__init__.py
+++ b/deepgram/__init__.py
@@ -14,7 +14,7 @@ from .errors import DeepgramSetupError, DeepgramApiError
 
 
 def api_key_is_valid(api_key: str) -> bool:
-    pattern = r"^[a-z0-9]{40}$"
+    pattern = r"^[a-f0-9]{40}$"
     return re.match(pattern, api_key) is not None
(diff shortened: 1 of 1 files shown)


deepgram/__init__.py
```diff
@@ -12,6 +13,11 @@
 from .errors import DeepgramSetupError, DeepgramApiError
 
 
+def api_key_is_valid(api_key: str) -> bool:
+    pattern = r"^[a-z0-9]{40}$"
```
jkroll-deepgram, line 17: I think it is only `a-f`, not `a-z` - is that too finicky? I think fine as-is, it will do the trick as it's an odd edge case for a user to provide a 40-char false API key.

jjmaldonis: It's easy enough to fix, thanks!

jkroll-deepgram (approved):

PR #137, opened 2023-08-23
adds examples folder with projects to do local testing

SandraRodgers:
This PR does the following:
- Adds an 'examples' folder with two projects that are meant to be used to do local testing

The purpose of this is to add examples that can be used to more easily test local changes to the SDK when doing work to the SDK.


tests/test_billing.py
jjmaldonis, line 1: This is a side note, but I don't understand why Saarth's PR didn't make it into the repo: https://github.com/deepgram/deepgram-python-sdk/pull/134/files

The PR got merged but I don't see the file/code he wrote in the repo.

README.md
```diff
@@ -131,31 +131,38 @@ Depending on your preference, you can also add parameters as named arguments, in
...
 
-In the `sample-projects` folder, there are examples from four different Python web frameworks of how to do live streaming audio transcription with Deepgram. These include:
+### Setup
 
-- Flask 2.0
-- FastAPI
-- Django
-- Quart
+```
+pip install pytest
```
jjmaldonis, line 139: We have a [requirements-dev.txt](https://github.com/deepgram/deepgram-python-sdk/blob/main/requirements-dev.txt) file in the repo that includes `pytest` and some other dependencies, so we can tell the user to run `pip install -r requirements-dev.txt` instead.

`pytest-cov` needs to be added to that file.

README.md
```diff
@@ -131,31 +131,38 @@ Depending on your preference, you can also add parameters as named arguments, in
...
+pip install pytest-cov
+```
 
-## Testing
+### Run All Tests
 
 ```
 pytest --api-key <key> tests/
```
jjmaldonis, line 146: I would love it if I didn't have to pass the API key in on the command line to run the tests. If the argument is missing, can we default to having the code check the environment variable `DEEPGRAM_API_KEY`?

This change doesn't need to be done in this PR (or at all if other people feel differently), so feel free to resolve this comment with no changes if you want.

examples/README.md
```diff
@@ -0,0 +1,61 @@
+# Examples for Testing Features Locally
+
+The example projects are meant to be used to test features locally by contributors working on this SDK. In the `requirements.txt` file, the deepgram package used is the local version:
+
+```
+-e ../../
```
jjmaldonis, line 6: What is this code snippet referencing? Is there a package version in the `../..` directory somewhere?

examples/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+`DEEPGRAM_API_KEY = 'YOUR_DEEPGRAM_API_KEY'`
+
+### Run the project
+
+Make sure you're in the directory with the `main.py` file and run the project with the following command.
+
+`python3 main.py`
+
+## How to verify that you're testing the local changes
```
jjmaldonis, line 34: It may be useful to explain to the user how to install the Deepgram SDK from a local repo. We can do that using:

```bash
pip uninstall deepgram-sdk  # If it's already installed
cd /path/to/deepgram-python-sdk/
pip install -e .
```

This will install the SDK from the user's local source code, which allows them to make changes that are instantly usable in their local examples.

examples/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+Replace the API key where it says 'YOUR_DEEPGRAM_API_KEY'
+
+`DEEPGRAM_API_KEY = 'YOUR_DEEPGRAM_API_KEY'`
+
+### Run the project
+
+Make sure you're in the directory with the `main.py` file and run the project with the following command.
+
+`python3 main.py`
```
jjmaldonis, line 32: I really hope no one is using python 2 anymore 👀  I think we can change this to `python main.py`

examples/README.md
```diff
@@ -0,0 +1,61 @@
...
+If you are contributing changes to this SDK, you can test those changes by using the `prerecorded` or `streaming` projects in the `examples` folder. Here are the steps to follow:
+
+### Add Your Code
+Make your changes to the SDK (be sure you are on a branch you have created to do this work).
+
+### Install dependencies
+
+Run the following command to install the project dependencies:
+
+`pip install -r requirements.txt`
```
jjmaldonis, line 20: Does the user have to be in a specific repo to run this command?

examples/prerecorded/main.py
```diff
@@ -0,0 +1,72 @@
+# Example filename: deepgram_test.py
+
+import json
+import asyncio
+
+from deepgram import Deepgram
+
+# Your Deepgram API Key
+# DEEPGRAM_API_KEY = 'YOUR_DEEPGRAM_API_KEY'
+DEEPGRAM_API_KEY = ''
```
jjmaldonis, line 10: Maybe delete this line and just keep the one above?

examples/README.md
```diff
@@ -0,0 +1,61 @@
+# Examples for Testing Features Locally
+
+The example projects are meant to be used to test features locally by contributors working on this SDK. In the `requirements.txt` file, the deepgram package used is the local version:
```
jjmaldonis, line 3: I don't exactly understand the purpose of this `examples` code.

Let's say I'm making a change to the SDK and I'm adding a new feature, `filler_words` to the SDK. I know that I need to add the `filler_words` string to the Options dictionary, so I make those changes.

Now I need to test those changes, so I modify `examples/prerecorded/main.py` to include the `filler_words=true` option. I run `python main.py` and it works!!!

Now I want to commit my additions to the SDK. Do I commit `examples/prerecorded/main.py`?

I'm guessing the answer is "no", and that I should write a unit test instead. If that's the case, I think we want to rename the `examples` directory to something like `testing_examples` to make it clear this code is not meant to contain examples for the user to start their project.

I also think the user _can_ use this code to start their own project. If that's the case, then I think we rewrite this sentence to explain these `prerecorded` and `streaming` projects are basic examples, and they _should_ be used to test contributions to the SDK. -- Maybe this isn't the right thing to do, but I don't understand what exactly this code is supposed to be used for / how it should be used.

jjmaldonis (changes requested):

examples/README.md
SandraRodgers: When I researched how we could use a local file for the deepgram package (instead of installing the package the normal way), I learned that we could add `-e ../../` to the requirements.txt file, creating the editable reference to the local deepgram package. 

The way I understand it is the -e flag indicates editable mode, meaning that the package is being installed from the directory in a way that any changes made to the code are immediately reflected without needing to reinstall. Unlike traditional package installations, which often specify specific versions, we're using the editable reference to allow us to work on and test the deepgram package directly within our project....

Now what I'm not sure is, based on your other comment, is this the best way to do this? We want to have folder like this in all the SDKs which has a pre-recorded and a streaming file separate from the SDK. The idea is to use it to test local changes. 

Do you want this to be changed to the method you suggested below where we do this:

```
pip uninstall deepgram-sdk  # If it's already installed
cd /path/to/deepgram-python-sdk/
pip install -e .
```

I want to do this in a way that makes sense to people so please let me know if you think I should just scrap this method

examples/README.md
SandraRodgers: So this means the person would uninstall the sdk, then go into the local deepgram sdk (I guess at the root of the project) and do the installation again with the -e flag. So the -e flag is what makes it install the local version?

examples/README.md
SandraRodgers: > Now I want to commit my additions to the SDK. Do I commit examples/prerecorded/main.py?

Correct, the answer is "no". The reality is that sometimes we need to be able to try things out quickly. The example files allow us to make quick changes to the SDKs and test them out within the same repo. We're adding these to all the SDKs.

I don't have a strong opinion about which features are the permanent ones in the `examples/prerecorded/main.py` as the example options to test. But the returned data should be JSON in the terminal so someone can quickly see if the response looks the way it should.

I really hope we can get to a point where people are writing unit tests with every new feature. We're moving in that direction but for now we're still at the stage where we're trying to get more people to actually just test the SDKs thoroughly when people do code reviews.

README.md
SandraRodgers: Should I remove the instructions to `pip install pytest-cov` and `pip install pytest` since these are in the requirements-dev.txt file?

README.md
SandraRodgers: I'll have to look into making it work that way. No promises. 🙂

examples/README.md
SandraRodgers: Yes, in the examples folder in either the prerecorded or streaming project. I'll add that clarification.

tests/test_billing.py
SandraRodgers: Actually I think it is there! I don't know what's going on here.

examples/README.md
jjmaldonis: Yeah exactly - it's a really nice way to use packages that you've created (or are editing) locally. The package needs to have a `setup.py` file, which is also required for pypi and the Deepgram SDK has it already.

README.md
jjmaldonis: Yeah, you could replace those two `pip install` lines with `pip install -r requirements-dev.txt`

examples/README.md
jjmaldonis: Interesting, I didn't understand what the `-e ../../` was doing. So yeah, it's doing the same thing as `pip install -e .`. I don't have a strong preference for which way we tell users how to install their local version of the Deepgram SDK. If we keep the `-e ../../` in the `requirements.txt` file, can we add a comment in the Readme explaining that running `pip install requirements.txt` will install the package itself? Typically running `pip install requirements.txt` installs the _dependencies_ for the package rather than the package itself, so we'd just want to document what's going on since it's a little out of the ordinary (but still okay if you want to go that route imo).

examples/README.md
jjmaldonis: Gotcha, that makes sense now. It might be worth a sentence in the Readme saying that the code changes the person makes to the files in `examples/...` can be discarded when they commit their SDK updates.

PR #138, opened 2023-08-24
chore: add pre-commit to check conventional commit msg

Weimw:
Add pre-commit to check Conventional Commits(https://www.conventionalcommits.org/en/v1.0.0/).

To finish setup:  

1.  Run `pip install -r requirements-dev.txt`
2. Run `pre-commit install --hook-type commit-msg `


dvonthenen:
I think work on this has stalled. If that isn't the case, this can be reopened.

Issue #158, opened 2023-10-24
Remove MIME type for Batch


SandraRodgers:
## Proposed changes

Our ASR API never requires a MIME type unless you are making a fetch request with a JSON body, in which case it must be application/json.

## Context

NA

## Possible Implementation

NA

## Other information

NA

dvonthenen:
PR merged. Closing as fixed.

PR #172, opened 2023-11-09
Restructure Python v3 (Part 2) - Update Project Settings

dvonthenen:
This builds on this PR https://github.com/deepgram/deepgram-python-sdk/pull/156 to restructure the project. The goal is to do this in smaller chunks so the work is not finished in this PR. The PR merges to main so the PR is going to look messy until the dependent PR is merged.

Here is a clean diff in the meantime: https://github.com/deepgram/deepgram-python-sdk/compare/sr/restructure-files...dvonthenen:deepgram-python-sdk:restructure-project-part2

This PR addresses the following issue(s): https://github.com/deepgram/deepgram-python-sdk/issues/158
Verified that `examples/prerecorded` (and all demos in examples) still works.

Notable changes:
- restructure `deepgram/__init__.py` so that it conforms to the usual python layout. `__init__.py` is now just the landing point for importing all things
- moved `__version__` (thereby deleting `_version.py`) into `__init__.py` using newer features in `setuptools` package
- removed `mimetypes` in prerecorded API calls
- Removed `TranscriptionOptions` in `transcription_options.py` which just looks like dead code
- Explicitly made use of the input `PrerecordedOptions ` and `LiveOptions`. This is mainly for user understanding.


.env
```diff
@@ -0,0 +1,8 @@
+DG_API_KEY="ecec76b6cd2c0b75d45500734c4fb37e2be6f09e"
```
SandraRodgers, line 1: Hmmm... why did this get pushed up? I see .env in the .gitignore file. 

I'll kill all these api keys. We should delete this

SandraRodgers (approved):
There are some __pycache__ files that were pushed to the branch. I don't think we want those in github, do we?

.env
dvonthenen: I noticed it later on in one of the other PRs and removed it there. This should get taken care of down the line.

deepgram-devrel (approved):

jpvajda (approved):

PR #184, opened 2023-11-20
adds captions library and example

SandraRodgers:
This adds dependency for deepgram-captions. There is an example file and a readme update to show how to create SRT and webVTT captions from the SDK.

### DIFF
--- a/examples/captions/main.py
+++ b/examples/captions/main.py
@@ -1,0 +1,55 @@
+# Copyright 2023 Deepgram SDK contributors. All Rights Reserved.
+# Use of this source code is governed by a MIT license that can be found in the LICENSE file.
+# SPDX-License-Identifier: MIT
+
+import asyncio
+import os
+import json
+from dotenv import load_dotenv
+from deepgram_captions import DeepgramConverter, webvtt, srt
+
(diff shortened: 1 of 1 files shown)


dvonthenen (approved):

dvonthenen (commented):
Removed

examples/captions/main.py
```diff
@@ -0,0 +1,55 @@
+# Copyright 2023 Deepgram SDK contributors. All Rights Reserved.
+# Use of this source code is governed by a MIT license that can be found in the LICENSE file.
+# SPDX-License-Identifier: MIT
+
+import asyncio
+import os
+import json
+from dotenv import load_dotenv
+from deepgram_captions import DeepgramConverter, webvtt, srt
```
dvonthenen, line 9: Wouldn't this go in the captions repo under its own examples folder?

dvonthenen:
We had discussed this, including this would create a circular dependency. Installing the captions package wouldn't turn install the deepgram SDK based on the version it defined in its requirements/setup definitions.

You can find an example of a captions implementation in source in the repo:
https://github.com/deepgram/deepgram-python-captions/blob/main/example.py

PR #186, opened 2023-11-22
Refactor Live Client for v3

dvonthenen:
This PR introduces the following:
- Refactor the Live client to use IO threading versus async/wait. Found a pretty significant issue which could affect throughput of streaming to the websocket. async/wait is good for non-real-time things as it simplifies things since async/wait uses a single thread for all functions, objects, etc. This turns out to be not so great for IO intensive tasks since pushing the IO will be interrupted from every other thread doing async/wait stuff because they all share a single thread. More info [here](https://docs.python.org/3/library/threading.html) and [here](https://realpython.com/async-io-python/).
- Preserves the async/await Live client implementation to be marked for deprecation in the next major/minor version. A warning will be printed to the user informing them this will be deleted in a future release.
- Introduces a helper class to implement microphone input using pyaudio (backed by portaudio).
- Introduces a simple example "main" style application (located at `examples/streaming/microphone`) of using the microphone with the Live client.
- Refactors the errors not be global and places them into their own modules.


deepgram/clients/live/v1_client.py
```diff
@@ -40,66 +43,148 @@ def __init__(self, base_url, api_key, headers):
...
+        try:
+            time.sleep(PING_INTERVAL)
+
+            self.lock_exit.acquire()
+            myExit = self.exit
+            self.lock_exit.release()
+            if myExit:
+                return
+
+            # deepgram keepalive
```
lukeocodes, line 143: I didn't think we would be introducing the default KeepAlive until we had discussed whether it breaks billing on lag. Did we get an answer?

dvonthenen: I thought this was to be included. If not, I can no longer make it the default and put it behind an option flag.

dvonthenen: Since this is early on (still dont have an alpha yet), I will follow up with that on a subsequent PR. I want to get an alpha out the door ASAP.

SandraRodgers (approved):

deepgram/clients/live/v1_client.py
dvonthenen: The PR was updated to disable by default but optional be enabled

PR #200, opened 2023-12-07
Implement Sync (Default) and Async For All Functions/Classes

davidvonthenen:
Depends on https://github.com/deepgram/deepgram-python-sdk/pull/199

This implements Sync (Default) and Async For All Functions/Classes. I was originally going to deprecate the Async to reduce the amount of code to maintain, but I convinced myself to support both. There is some utility to having the Async functions as they are good for things like:
- quick development and prototyping
- demos when speed isn't important

All examples were updated to use the sync functions/classes, and provided a couple of examples for async.

This follows the pattern a lot of other IO/CPU intensive libraries have set where sync/threading is the default and async is an option. Some good examples to reference are libraries that we consume in this SDK. For example:
- WebSockets
  - threading is the default - https://github.com/python-websockets/websockets/blob/main/src/websockets/client.py
  - async is legacy - https://github.com/python-websockets/websockets/blob/main/src/websockets/legacy/client.py
- httpx
  - threading/sync is default:
    ```
    >>> import httpx
    >>> r = httpx.get('https://www.example.org/')
    >>> r
    <Response [200 OK]>
    ```
  - async requires an additional property for access:
    ```
    >>> async with httpx.AsyncClient() as client:
    ...     r = await client.get('https://www.example.com/')
    ...
    >>> r
    <Response [200 OK]>
    ```

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,29 @@
+# Examples for Testing Features Locally
+
+The example projects are meant to be used to test features locally by contributors working on this SDK.
+
+## Steps to Test Your Code
+
+If you are contributing changes to this SDK, you can test those changes by using the `prerecorded`, `streaming`, or `manage` "hello world"-style applications in the `examples` folder. Here are the steps to follow:
+
+### Set your API Key as an Environment Variable named "DEEPGRAM_API_KEY"
+
+If using bash, this could be done in your `.bash_profile` like so:
+
+```bash
+export DEEPGRAM_API_KEY = "YOUR_DEEPGRAM_API_KEY"
+```
+
+or this could also be done by a simple export before executing your python application:
+
+```bash
+DEEPGRAM_API_KEY="YOUR_DEEPGRAM_API_KEY" python main.py
+```
+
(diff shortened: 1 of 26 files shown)


dvonthenen:
rebased changes after merging https://github.com/deepgram/deepgram-python-sdk/pull/199

SandraRodgers (approved):

examples/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+or this could also be done by a simple export before executing your python application:
+
+```bash
+DEEPGRAM_API_KEY="YOUR_DEEPGRAM_API_KEY" python main.py
+```
+
+### Run the project
+
+Make sure you're in the directory with the `main.go` file and run the project with the following command.
```
SandraRodgers, line 25: need to change to python

SandraRodgers (approved):

examples/README.md
dvonthenen: ooops I forgot to save the file as second time before committing it. should be fixed now.

PR #221, opened 2023-12-22
Update README for SemVer Guidance

dvonthenen:
This explains semver best-practices and guarantees about major interface changes specifically around major version releases (ie `v2` to `v3).


lukeocodes (approved):
Nice

PR #224, opened 2023-12-23
Update README.md - Firm Min Support

dvonthenen:
Looks like 3.9 or lower won't work out of the box. Will need to explore if it's worth it. Until then...

3.10 has been around for a while (since 2020) and is in maintenance at this point.

3.8 will be EOL this year, 3.9 will be EOL next year. https://devguide.python.org/versions/

The min version (without going through heroic efforts) is a firm 3.10.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,8 +31,6 @@ You can learn more about the Deepgram API at [developers.deepgram.com](https://d
 
 [Python](https://www.python.org/downloads/) (version ^3.10)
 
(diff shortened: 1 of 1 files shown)


dvonthenen:
Tried using latest latest 3.9.16 I think... and same. Looks like Python (rightly so) doesn't introduce new features until the next minor release.

SandraRodgers (approved):

Issue #170, opened 2023-11-06
Manage endpoints updated to support Tokens 


jpvajda:
Manage endpoints that will likely change:

* `request` - https://developers.deepgram.com/reference/get-request
* `requests`  - https://developers.deepgram.com/reference/get-all-requests
* `usage` - https://developers.deepgram.com/reference/summarize-usage
* `fields` - https://developers.deepgram.com/reference/get-fields"

UniversalDevGenius:
@jpvajda I would like to work on this issue, can you provide more context regarding it.

dvonthenen:
@UniversalDevGenius You missed it by about a day... it's mostly implemented here:
https://github.com/dvonthenen/deepgram-python-sdk/tree/feature-sits-impl

Just need to rename a couple of functions after discussing what they should be called.

UniversalDevGenius:
> @UniversalDevGenius You missed it by about a day... it's already implemented here: https://github.com/dvonthenen/deepgram-python-sdk/tree/feature-sits-impl
> 
> Just need to rename a couple of functions after discussing what they should be called.

Can you share me some issues that I can contribute just tag them here or assign them to me, I really appreciate if you give me an chance to work on interesting issue to start mine first every open source contribution.

UniversalDevGenius:
> hey @UniversalDevGenius drop me a line in Discord (https://dpgr.am/discord) if you are interested and we can discuss. The APIs for 3.1 are mostly finalized and the deadline is fast approaching.

Responded to you on Discord as well. Thanks for responding. Looking forward to discuss more on it.

dvonthenen:
I think this is mostly implemented and was available in `v3.0.0`... might need to investigate some:
https://github.com/deepgram/deepgram-python-sdk/blob/main/deepgram/clients/manage/v1/client.py#L621-L747

dvonthenen:
Looks like new fields on the response objects. That's about the extent of the work.

dvonthenen:
@UniversalDevGenius 

for the support endpoint tokens (https://github.com/deepgram/deepgram-python-sdk/issues/170), I will double check... but you should just need to enable the debug flag in the example (https://github.com/deepgram/deepgram-python-sdk/blob/main/examples/manage/usage/main.py) and then grab a copy of the response JSON (in the debug statements it will be the output starting with `json:`) and then run the json through https://jsonformatter.org/json-to-python to copy over the new fields in the response objects

[...]

PR #259, opened 2024-01-17
 modified ` response.py ` to add support for tokens issue #170

UniversalDevGenius:
add new fields related to token support that resulted in modification of ` response.py ` , these changes related to issue #170


dvonthenen (approved):

dvonthenen:
Sweet! Thanks for your contribution @UniversalDevGenius. This is going to be super helpful for those individuals looking to understand how tokens are consumed through the various APIs!

Issue #282, opened 2024-02-01
Calling `PreRecordedClient.transcribe_url_callback` with initialized `options` results in `TypeError`


voyc-cameron:
## What is the current behavior?

I have encountered an issue while attempting to migrate our current Deepgram interface from `requests` to the SDK.

A `TypeError` occurs when making a prerecorded callback request with an `options` object. 
The method's type hints says it expects `options` to be an instance of `PrerecordedOptions`. 
However, the code attempts to set the `callback` attribute as an item rather than an attribute.

```python
# deepgram/clients/prerecorded/v1/client.py

    def transcribe_url_callback(
        self,
        source: UrlSource,
        callback: str,
        options: PrerecordedOptions = None,
        addons: dict = None,
        timeout: httpx.Timeout = None,
        endpoint: str = "v1/listen",
    ) -> AsyncPrerecordedResponse:
        self.logger.debug("PreRecordedClient.transcribe_url_callback ENTER")

        url = f"{self.config.url}/{endpoint}"
        if options is None:
            options = {}
        options["callback"] = callback  # <- error occurs here
```

The issue is further compounded as if `options` is `None` then it would be initialised as a dictionary (which works for initially setting `callback`); however, further down on [line 131](https://github.com/deepgram/deepgram-python-sdk/blob/e2ddfb4f97def5e5c71910bf2a84b50c8a4ba349/deepgram/clients/prerecorded/v1/client.py#L131) the method attempts to call `options.check()` which fails - so passing a `options` dictionary to the method does not work either.

The same issue above occurs whether a callback request is made via calling `.transcribe_url` which indirectly calls `.transcribe_url_callback`. I also noticed that this behavior exists in `AsyncPreRecordedClient` too.

I'm not too sure if this is suppose to be expected behavior from utilizing `dataclass_json`; however, I couldn't find any item setter logic there. 

It would probably be best if the variables adhere to the type hints and `options` is initialized as an object if it is indeed `None`, i.e.
```python
        if options is None:
            options = PrerecordedOptions()
        options.callback = callback
```

## Steps to reproduce

```python
AUDIO_URL = {
    "url": "https://static.deepgram.com/examples/Bueller-Life-moves-pretty-fast.wav"
}

# STEP 1 Create a Deepgram client using the API key from environment variables
deepgram = DeepgramClient()

# STEP 2 Call the transcribe_url method on the prerecorded class
callback_url = "https://example.com/callback"
options = PrerecordedOptions(
[...]

PR #285, opened 2024-02-02
Fix for `transcribe_url_callback` and `transcribe_file_callback` with Example

dvonthenen:
## Proposed changes

This addresses issue: https://github.com/deepgram/deepgram-python-sdk/issues/282

Fixes a couple of issues with `transcribe_url_callback` and `transcribe_file_callback` both in setting the callback field as well as using the incorrect type for the `httpx` when calling `post` (`json` vs `content`).

Also provided an example to demonstrate the fix.

## Types of changes

## Checklist

## Further comments

NA


examples/prerecorded/callback/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+The `callback` component requires the Deepgram API Key environment variable to be configured to run.
+
+```sh
+export DEEPGRAM_API_KEY=YOUR-APP-KEY-HERE
+```
+
+## Installation
+
+Run the `endpoint` application using an SSL certificates to a system on the public-facing internet. This could be an EC2 instance on AWS, an instance on GCP, or etc.
```
jkroll-deepgram, line 21: Nit: `an SSL certificates` - just one certificate? certificate~~s~~

examples/prerecorded/callback/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+On the `callback` project, modify the IP address constant in the code with the public-facing IP address of your EC2, GCP, etc instance.
```
jkroll-deepgram, line 23: Can a user run this locally too? Having another server is an extra lift to get started with this example.

examples/prerecorded/callback/callback/main.py
```diff
@@ -0,0 +1,66 @@
...
+            utterances=True,
+            punctuate=True,
+        )
+
+        response = deepgram.listen.prerecorded.v("1").transcribe_file_callback(
+            payload, CALL_BACK_URL, options=options
+        )
+        # response = deepgram.listen.prerecorded.v("1").transcribe_url_callback(
+        #     payload, CALL_BACK_URL, options=options
+        # )
```
jkroll-deepgram, line 58: Can delete this commented-out duplicate block?

examples/prerecorded/callback/callback/main.py
```diff
@@ -0,0 +1,66 @@
...
+        }
+        # payload: UrlSource = {
+        #     "url": "https://static.deepgram.com/examples/Bueller-Life-moves-pretty-fast.wav"
+        # }
+
+        options = PrerecordedOptions(
+            model="nova-2",
+            smart_format=True,
+            utterances=True,
+            punctuate=True,
```
jkroll-deepgram, line 50: General parameters feedback: My understanding is that `punctuate` is a subset of `smart_format`, so both aren't needed in the same API call. `punctuate` is when you only want punctuation without the rest of entity/numerals/etc formatting.

examples/prerecorded/callback/README.md
dvonthenen: updated!

examples/prerecorded/callback/README.md
dvonthenen: They can... they just need a public-facing IP address since the DG platform needs to actually hit your endpoint. You can do this by setting up port-forwarding on your router (assuming your ISP won't block/filter hosting a server).

I think someone who knows anything about networking should know how to do this, but if you have any ideas on what to say to someone who doesn't have that networking experience, I would be open to updates.

examples/prerecorded/callback/callback/main.py
dvonthenen: I put that there so people using this can also do this without using a local file.

examples/prerecorded/callback/callback/main.py
dvonthenen: Copy and paste error. Will fix!

examples/prerecorded/callback/README.md
jkroll-deepgram: `ngrok` is the go-to for simple callback testing.

Brief overview: run `ngrok http 8080` which opens up the local host port 8080, and gives a public URL `https://e00a-24-165-89-177.ngrok-free.app/` (for example).  Then in another terminal, run `nc -l 8080` to listen on port 8080 for incoming messages.

The request url is then `https://api.deepgram.com/v1/listen?model=general&callback=https://e00a-24-165-89-177.ngrok-free.app`.  The request immediately gets a 200, and the nc terminal eventually gets the Deepgram callback response JSON printed out.

Not saying it's strictly necessary to incorporate this into your example, but I do think it's worth mentioning somewhere to provide a lightweight local callback option!

dvonthenen: Sounds good. Will add that in.

lukeocodes (approved):

PR #355, opened 2024-04-05
fix: handle socket connection closed error in _signal_exit

roerohan:
## Proposed changes

In the `_listening` function, when there's a `WebSocketException`, the following line runs as a part of the error handling procedure.

https://github.com/deepgram/deepgram-python-sdk/blob/d2c43344e723df99468bff161e5e2a1c29600fd8/deepgram/clients/live/v1/async_client.py#L371

Now, the `WebSocketException` might also be caused due to the socket connection breaking. In this case, an error is thrown of the following form

```
future: <Task finished name='Task-291' coro=<AsyncLiveClient._listening() done, defined at /usr/local/lib/python3.12/dist-packages/deepgram/clients/live/v1/async_client.py:171> exception=ConnectionClosedError(Close(code=1011, reason='Deepgram did not receive audio data or a text message within the timeout window. See https://dpgr.am/net0001'), Close(code=1011, reason='Deepgram did not receive audio data or a text message within the timeout window. See https://dpgr.am/net0001'), True)>
```

While handling this exception, the `_signal_exit` function is called, which has the following lines
https://github.com/deepgram/deepgram-python-sdk/blob/d2c43344e723df99468bff161e5e2a1c29600fd8/deepgram/clients/live/v1/async_client.py#L483-L496

In line 488, we're doing a `self._socket.send`, which is bound to fail because the socket connection is closed. This throws an error from the `_signal_exit` method. Since `_signal_exit` is not wrapped in a `try except`, the `_listening` method also throws an error and exits. The following is a sample traceback.

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/dist-packages/deepgram/clients/live/v1/async_client.py", line 287, in _listening
    await self._signal_exit()
  File "/usr/local/lib/python3.12/dist-packages/deepgram/clients/live/v1/async_client.py", line 478, in _signal_exit
    await self._socket.send(json.dumps({"type": "CloseStream"}))
  File "/usr/local/lib/python3.12/dist-packages/websockets/legacy/protocol.py", line 635, in send
    await self.ensure_open()
  File "/usr/local/lib/python3.12/dist-packages/websockets/legacy/protocol.py", line 939, in ensure_open
    raise self.connection_closed_exc()
```

[...]

PR #357, opened 2024-04-05
Implement Unhandled Events

dvonthenen:
## Proposed changes

This implements an `unhandled` event type. If there is a new event type encountered that the SDK currently doesn't implement, this will fire off a "unhandled' event type and just return the raw JSON to the user through this event. The idea is that the user can take this raw JSON and handle the event themselves until the event is supported in the SDK.

This is already implemented in the Go SDK:
https://github.com/deepgram/deepgram-go-sdk/blob/main/pkg/api/live/v1/interfaces/interfaces.go#L18

Also a slight addition for this PR that was submitted by a member in the community that was just merged:
https://github.com/deepgram/deepgram-python-sdk/pull/355

For the fix above, the emit close event needs to be before triggering the disconnect on the `set()` function. Otherwise, the event will not fire.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have lint'ed all of my code using repo standards

### DIFF
--- a/deepgram/__init__.py
+++ b/deepgram/__init__.py
@@ -23,8 +23,9 @@ from .client import (
     MetadataResponse,
     SpeechStartedResponse,
     UtteranceEndResponse,
-    ErrorResponse,
     CloseResponse,
+    UnhandledResponse,
+    ErrorResponse,
 )
 
 # prerecorded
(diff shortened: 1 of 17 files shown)


jpvajda (approved):
LGTM

PR #365, opened 2024-04-16
Handle NoneType exit event

jkroll-deepgram:
## Proposed changes

Fix a bug where `is_set()` is being called on `self._exit_event` when it has a value of `None`, causing an `AttributeError` that bubbles up as an error in application code. To fix, require that `self._exit_event is not None` before calling `self._exit_event.is_set()`.

I do also see in other spots that events are being initialized as `asyncio.Event()`. Should the exit event be instantiated as `self._exit_event = asyncio.Event()` instead of `self._exit_event = None`, so `self._exit_event` is never `None` in the first place?

## Types of changes

## Checklist

- [ ] I have read the [CONTRIBUTING](../CONTRIBUTING.md) doc

### DIFF
--- a/deepgram/clients/live/v1/async_client.py
+++ b/deepgram/clients/live/v1/async_client.py
@@ -184,7 +184,7 @@ class AsyncLiveClient:
 
         while True:
             try:
-                if self._exit_event.is_set():
+                if self._exit_event is not None and self._exit_event.is_set():
                     self.logger.notice("_listening exiting gracefully")
                     self.logger.debug("AsyncLiveClient._listening LEAVE")
                     return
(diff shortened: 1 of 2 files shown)


dvonthenen (approved):

dvonthenen:
> I do also see in other spots that events are being initialized as asyncio.Event(). Should the exit event be instantiated as self._exit_event = asyncio.Event() instead of self._exit_event = None, so self._exit_event is never None in the first place?

I can take a look at this and double check.

dvonthenen:
I suspect that the event should have been reset and not deleted is the problem. This might have been a copy and paste gone too far. Confirming this now.

PR #400, opened 2024-05-17
Change utt_split type to float

RyanChimienti:
The [docs](https://developers.deepgram.com/docs/utterance-split#enable-feature) show that `utt_split` is actually a float value, so I think this type should be changed.

I tested this change by printing out `_url` in `abstract_sync_client.py > _handle_request()`, and then making Prerecorded requests with and without this change. For `utt_split` values of `0.001`, `2`, and `1.01000`, the `_url` was the same with and without this change.

In other words, this change seems to be a functional no-op and is just documentation.


dvonthenen:
Yep, that's an oops. Nice catch!

PR #419, opened 2024-06-13
Stop importing pyaudio as soon as the Microphone class is parsed

### LINKED ISSUE #418: v3.3.0 needs pyaudio but doesn't specify it as a dependency
## What is the current behavior?

Installing v3.3.0 of this library and trying to `import deepgram` gives a stack trace because deepgram-sdk depends on pyaudio, but it isn't installed as a dependency.

## Steps to reproduce

1. Create an empty virtual environment: `python3 -m venv venv`
2. Install deepgram-sdk v3.3.0 in the virtual environment: `venv/bin/pip install deepgram-sdk==3.3.0`
3. Try to import the library: `venv/bin/python -c 'import deepgram'`

It gives a stack trace like this:
```
  File "<string>", line 1, in <module>
  File "/your/path/venv/lib/python3.12/site-packages/deepgram/__init__.py", line 129, in <module>
    from .audio import Microphone
  File "/your/path/venv/lib/python3.12/site-packages/deepgram/audio/__init__.py", line 5, in <module>
    from .microphone import Microphone
  File "/your/path/venv/lib/python3.12/site-packages/deepgram/audio/microphone/__init__.py", line 5, in <module>
    from .microphone import Microphone
  File "/your/path/venv/lib/python3.12/site-packages/deepgram/audio/microphone/microphone.py", line 15, in <module>
    class Microphone:  # pylint: disable=too-many-instance-attributes
  File "/your/path/venv/lib/python3.12/site-packages/deepgram/audio/microphone/microphone.py", line 23, in Microphone
    import pyaudio  # pylint: disable=import-outside-toplevel
    ^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'pyaudio'
```

## Expected behavior

It should be possible to install the library and import it without having to install extra dependencies manually.

## Please tell us about your environment

This happens regardless of operating system. I have only tested on Python 3.12.

Tenzer:
## Proposed changes

The import is only used for type checking, so it can instead be put inside an `if TYPE_CHECKING:` block so the optional dependency isn't needed at all times.

## Types of changes

## Checklist

- [ ] I have read the [CONTRIBUTING](../CONTRIBUTING.md) doc

## Further comments

Fixes #418.

### DIFF
--- a/deepgram/audio/microphone/microphone.py
+++ b/deepgram/audio/microphone/microphone.py
@@ -5,11 +5,14 @@
 import inspect
 import asyncio
 import threading
-from typing import Optional, Callable
+from typing import Optional, Callable, TYPE_CHECKING
 import logging
 
 from ...utils import verboselogs
 from .constants import LOGGING, CHANNELS, RATE, CHUNK
+
+if TYPE_CHECKING:
+    import pyaudio
 
 
 class Microphone:  # pylint: disable=too-many-instance-attributes
@@ -20,10 +23,8 @@ class Microphone:  # pylint: disable=too-many-instance-attributes
     _logger: verboselogs.VerboseLogger
     _exit: threading.Event
 
-    import pyaudio  # pylint: disable=import-outside-toplevel
-
-    _audio: pyaudio.PyAudio
-    _stream: pyaudio.Stream
+    _audio: "pyaudio.PyAudio"
+    _stream: "pyaudio.Stream"
     _chunk: int
     _rate: int
     _format: int


dvonthenen:
I am going to keep this PR open until I have time to investigate this. I think `TYPE_CHECKING` was introduced later in python v3 and need to check if it will work with the versions we support. This would be the ideal fix for this problem.

In the meantime, I am going to just revert and disable the static checking on this.

Tenzer:
It was added in Python 3.5.2, according to the documentation: https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING.

dvonthenen (approved):

dvonthenen:
@Tenzer thanks again for the PR. just posted a new release with this fix:
https://github.com/deepgram/deepgram-python-sdk/releases/tag/v3.3.1

Tenzer:
Thanks! I can see the new version passes tests in our application at least, so that's better than with version 3.3.0 :)

PR #534, opened 2025-05-31
fix: correct the language property location and type

jpvajda:
## Proposed changes

Relates to the bug reported in [The community ](https://github.com/orgs/deepgram/discussions/1067#discussioncomment-13326127)

Let me analyze what I found:

The Language class is defined in options.py but it's not being used in any of the main client classes or settings. It appears to be defined but not integrated into the actual SDK functionality.

Looking at the SettingsOptions class which is the main configuration class used when starting a websocket connection, it doesn't include the Language class as a field.

## Types of changes

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1210418536779480

## Summary by CodeRabbit

- **New Features**
  - Added support for specifying a language option in the agent configuration.
- **Bug Fixes**
  - Corrected language and model property assignments in example configurations for better clarity and structure.
- **Chores**
  - Simplified language field type in agent settings and cleaned up redundant code formatting.

### DIFF
--- a/deepgram/clients/agent/v1/websocket/options.py
+++ b/deepgram/clients/agent/v1/websocket/options.py
@@ -363,6 +363,7 @@ class SettingsOptions(BaseResponse):
     type: str = str(AgentWebSocketEvents.Settings)
     audio: Audio = field(default_factory=Audio)
     agent: Agent = field(default_factory=Agent)
+    language: Language = field(default_factory=Language)
 
     def __getitem__(self, key):
         _dict = self.to_dict()
@@ -370,6 +371,8 @@ class SettingsOptions(BaseResponse):
             _dict["audio"] = Audio.from_dict(_dict["audio"])
         if "agent" in _dict and isinstance(_dict["agent"], dict):
             _dict["agent"] = Agent.from_dict(_dict["agent"])
+        if "language" in _dict and isinstance(_dict["language"], dict):
+            _dict["language"] = Language.from_dict(_dict["language"])
         return _dict[key]
 
     def check(self):
--- a/deepgram/clients/agent/v1/websocket/options.py
+++ b/deepgram/clients/agent/v1/websocket/options.py
@@ -212,6 +212,13 @@ class SpeakProvider(BaseResponse):
             _dict["voice"] = CartesiaVoice.from_dict(_dict["voice"])
         return _dict[key]
 
+@dataclass
+class Language(BaseResponse):
+    """
+    Define the language for the agent.
+    """
+
+    type: str = field(default="en")
 
 @dataclass
 class Think(BaseResponse):
@@ -284,6 +291,7 @@ class Agent(BaseResponse):
     This class defines any configuration settings for the Agent model.
     """
 
+    language: Language = field(default_factory=Language)
     listen: Listen = field(default_factory=Listen)
     think: Think = field(default_factory=Think)
     speak: Speak = field(default_factory=Speak)
@@ -293,6 +301,8 @@ class Agent(BaseResponse):
 
     def __getitem__(self, key):
         _dict = self.to_dict()
+        if "language" in _dict and isinstance(_dict["language"], dict):
+            _dict["language"] = Language.from_dict(_dict["language"])
         if "listen" in _dict and isinstance(_dict["listen"], dict):
             _dict["listen"] = Listen.from_dict(_dict["listen"])
         if "think" in _dict and isinstance(_dict["think"], dict):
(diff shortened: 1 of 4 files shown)


deepgram/clients/agent/v1/websocket/options.py
```diff
@@ -363,13 +363,16 @@ class SettingsOptions(BaseResponse):
     type: str = str(AgentWebSocketEvents.Settings)
     audio: Audio = field(default_factory=Audio)
     agent: Agent = field(default_factory=Agent)
+    language: Language = field(default_factory=Language)
```
lukeocodes: John, I think language lives inside the agent object?

```
"agent": {
  "language": "en",
  "listen": {
    "provider": {
      "type": "deepgram",
      "model": "nova-3"
      // ... additional provider options: keyterms (nova-3 'en' only)
    }
  },
```

jpvajda: Good catch! I also saw the examples were all wrong using language so i fixed those in my recent commit.

deepgram/clients/agent/v1/websocket/options.py
```diff
@@ -284,6 +291,7 @@ class Agent(BaseResponse):
     This class defines any configuration settings for the Agent model.
     """
 
+    language: str = field(default="en")
```
lukeocodes, line 285: Small nit is that this is optional, but the default value will have virtually the same affect for the end user

jpvajda: @lukeocodes  Thanks! here's what I found: 

The field is technically optional (users don't have to provide it) But if they don't provide it, it will always be "en" due to the default value So in practice, the field will always have a value, either:

1. The value the user explicitly set
2. Or "en" if they didn't set anything

The feedback is pointing out that while making the field optional is technically correct according to the API spec, having a default value means that in practice, the field will always have a value. So the optionality doesn't really matter to the end user - they'll always get a value either way.

We could make it more explicitly optional by using:

`language: Optional[str] = field(default=None)`

But this would be less useful because:

The API will use "en" as the default anyway
It would require more null checks in the code
It doesn't provide any real benefit to the end user

Recommendation: Keep the current implementation with `default="en"` as it provides a better developer experience while still being technically correct.

deepgram/clients/agent/v1/websocket/options.py
```diff
@@ -212,6 +212,13 @@ def __getitem__(self, key):
             _dict["voice"] = CartesiaVoice.from_dict(_dict["voice"])
         return _dict[key]
 
+@dataclass
```
naomi-lgbt: We do not need the dataclass at all.

deepgram/clients/agent/v1/websocket/options.py
```diff
@@ -293,15 +301,15 @@ class Agent(BaseResponse):
 
     def __getitem__(self, key):
         _dict = self.to_dict()
+        if "language" in _dict and isinstance(_dict["language"], dict):
```
naomi-lgbt: It's a string field, so we do not need to instantiate the dataclass from a dict.

deepgram/clients/agent/v1/websocket/options.py
lukeocodes: This is correct but I err on the other side of the suggestion here. I try to leave as much handling as possible on the API. Reduces the footprint of the code

If you leave it as is, in the request logs they'll always be sending `language=en`, but if you change it to optional default=None, then it will rely on the API to default to English internally

So either way works for the end user. You own the change :)

deepgram/clients/agent/v1/websocket/options.py
jpvajda: @naomi-lgbt  Why is this the case for this option? We use dataclasses through the rest of the SDK:

and from my understanding this dataclass is not just a decorator - it's part of a larger pattern that provides type safety, validation, and consistent behavior across the SDK. Removing it would break this pattern and potentially introduce bugs or inconsistencies.

So if you can explain why it should be removed in this one case that would be ideal.

deepgram/clients/agent/v1/websocket/options.py
jpvajda: I see this comment, explains your above comment better.  So removing this and the data class is the right thing to do.  

Since language is defined as a string field with a default value of "en", there's no need to try to instantiate it as a Language dataclass. The current code is trying to handle a case that shouldn't exist - if language is a string field, it should never be a dictionary that needs to be converted to a Language object.

deepgram/clients/agent/v1/websocket/options.py
jpvajda: figured it out:

- The Language class is not exported or imported anywhere else
- The Agent class defines language as a simple string field with a default value of "en"
- The Language dataclass is not providing any additional functionality beyond what a string field already provides

jpvajda:
@naomi-lgbt  alrighty I think this should do it: 

The Language class has been successfully removed. The Agent class still has its language field defined as a simple string.

PR #712, opened 2026-05-12
fix: lowercase bool query params on websocket connect

GregHolmes:
## Summary
  
  Python's `urllib.parse.urlencode` stringifies booleans via `str()`, producing `True`/`False`
  (capitalized). The four websocket connect methods (`listen.v1`, `listen.v2`, `speak.v1`,
  `agent.v1`) all hand the encoded query dict to `urlencode`, so any `bool` kwarg — e.g.
  `diarize=True` — reaches the API as `?diarize=True` and is rejected with HTTP 400.

  The generated SDK types for these params are `Union[Literal["true","false"], Any]` because
  our AsyncAPI schemas declare them as string enums rather than `type: boolean`. The `Any` arm
  lets bools pass mypy without complaint, so the bug only surfaces on the wire. HTTP raw
  clients are unaffected because they hand params to httpx, which lowercases bools itself.

  ## Fix
  
  Add `_coerce_query_value` in `core/query_encoder.py` mapping `True`/`False` to `"true"`/
  `"false"` before `urlencode` runs. Applied at every scalar leaf site so bools nested in
  `list`/`dict` query values are also coerced. `isinstance(v, bool)` is precise — `1`/`0` are
  not coerced (regression test confirms).

  Not a breaking change: current behavior is "the call returns HTTP 400," so the fix can only
  turn broken calls into working ones. `diarize="true"` (string) still works identically.

  `query_encoder.py` is Fern-generated; the patch is frozen under the existing
  "temporarily frozen — manual patches" pattern in `.fernignore`, with the unfreeze condition
  documented in `AGENTS.md`.

  ## Test plan
  
  - [x] 12 new regression tests in `tests/custom/test_query_encoder.py`
  - [x] Full pytest suite: **197 passed, 4 skipped, 0 failed**
  - [x] Live before/after with `diarize=True` against production — unpatched returned HTTP
        400 (matching the reported bug); patched opened cleanly
  - [x] Six websocket-relevant examples completed end-to-end against production:
        `13` (listen v1), `14` (listen v2), `15` (batch HTTP with `diarize=True`),
        `21` and `24` (speak v1), `30` (agent v1)
  - [x] `ruff check` clean on touched files


lukeocodes (approved):
lgtm

PR #756, opened 2026-08-11
feat(telemetry): opt-out, isolated Sentry error telemetry for the SDK

GregHolmes:
## What

Adds an opt-out, isolated, error-only Sentry telemetry module (`src/deepgram/telemetry/`), wired into `DeepgramClient` / `AsyncDeepgramClient` via the existing `telemetry_opt_out` and `telemetry_handler` params. The goal is visibility into SDK failures (HTTP 5xx, transport/SDK errors) without touching customer data or the host app's own Sentry.

## Design

- **Isolated from the host app.** Never calls the global `sentry_sdk.init()`; builds a private `Client` with `default_integrations=False`, so we only ever capture SDK-originated errors and leave the customer's own Sentry setup untouched. This follows Sentry's own guidance for library authors.
- **Per-client, not global.** Capture routes through a per-client `TelemetrySink`, so one client's opt-out or custom handler can't leak into another's.
- **No personal data.** `send_default_pii=False`, `include_local_variables=False`, `max_request_body_size="never"`, plus a `before_send` scrubber that strips request bodies, query strings, cookies, credential headers, `server_name`, `user` (incl. geo), and redacts the home-dir username from stack-frame paths (`/Users/<name>/...` -> `/Users/<redacted>/...`).
- **Custom handler** fully replaces Sentry when supplied (nothing leaves the process).

## Safety / posture

- **Off by default and inert.** `telemetry_opt_out` defaults to `True`, `_EMBEDDED_DSN` is empty, and a `DEEPGRAM_TELEMETRY_DISABLED` kill-switch wins over everything. With no resolvable DSN this no-ops entirely, so **merging this ships nothing live** — arming is a separate, gated step.
- Load-shedding / bounded queue behaviour is inherited from `sentry-sdk`'s background worker (network post is off the request path).

## Testing

`tests/custom/test_telemetry.py` covers opt-out resolution, the env kill-switch, PII/secret scrubbing, stack-frame path redaction, per-client isolation, and host-app isolation. `pytest`, `ruff`, and `mypy` all pass.


paultill (changes requested):
Please see https://deepgram.slack.com/archives/C07E6NG16SZ/p1786549158800029?thread_ts=1772209500.054649&cid=C07E6NG16SZ

Sounds good. I'll state what I am hoping for concisely:

* For the first version, be as minimal as possible to keep complexity low. (Add more later once we de-risk the initial deployment of telemetry)
* Either modify Luke's doc or write your own doc to describe the PR's design. It should focus more on what you value rather than repeating the technical details that appear in the PR. Shorter is better!
* Be intentional about how you know that the new functionality in the SDK doesn't break client applications. This should be both testing (CC [@Edward Rosen](https://deepgram.slack.com/team/U0BDFQN4MBL)) and some design considerations such as bounded resource usage and making sure that errors posting telemetry are contained.

PR #750, opened 2026-07-31
feat(regen): listen v2 force-end-turn and redaction, listen v1 diarize metadata

GregHolmes:
Fern SDK regeneration for 2026-07-31.

**Status:** reconciled & verified — `pytest` 337 passed / 1 skipped · `ruff check src/deepgram` clean · `mypy` clean (879 files). **No breaking changes** — the public export surface has zero removals vs 7.6.0 (net +25 names).

## New in this regen
- **Listen V2 force-end-turn** — `send_force_end_turn()` + `ListenV2ForceEndTurn`, plus properly-typed `send_configure(ListenV2Configure)` and a `ListenV2ConfigureSuccess` response.
- **Listen V2 redaction** — `ListenV2Redact`.
- **Listen V1 diarize metadata** — `ListenV1ResponseMetadataDiarizeInfo` / `ListenV1ResultsMetadataDiarizeInfo` (+`Arch`).
- **Agent update-listen provider versioning** — `AgentV1UpdateListenListenProvider` (`_V1`/`_V2`, discriminated on `version`); new `ConversationHistoryMessageRole` / `FunctionCallHistoryMessageFunctionCallsItem`.
- **Client stream-reconnection config** — `stream_reconnection_enabled` / `max_stream_reconnection_attempts` on the client wrapper.
- **Dynamic wire version** — the SDK version reported in `User-Agent` / `X-Fern-SDK-Version` is now read from installed package metadata (`importlib.metadata`) instead of a hard-coded literal.

## Backward compatibility (all preserved — non-breaking)
- **Type renames aliased** (old import paths kept): `AnthropicThinkProviderModel`→`AnthropicModel`, `GoogleThinkProviderModel`→`GoogleModel`, `GroqThinkProviderReasoningMode`→`GroqReasoningMode`, `CartesiaSpeakProviderModelId`→`CartesiaModelId`, `CartesiaSpeakProviderVoice`→`CartesiaVoice`, `DeepgramSpeakProviderModel`→`DeepgramModel`.
- **Transitional agent-history names shimmed** — the 6 names the generator removed (`AgentV1HistoryContentRole`, `AgentV1HistoryFunctionCallsFunctionCallsItem` (+`Params`), and the double-`Context` `...ContentRole` / `...FunctionCallsFunctionCallsItem` (+`Params`)) are recreated as hand-written aliases to the new canonical types, so every old import path still resolves.
- **`send_configure`** still accepts a raw dict (sent verbatim) alongside the typed model.
[...]

PR #757, opened 2026-08-11
revert: listen v2 force-end-turn/redaction + diarize regen (#750)

GregHolmes:
Reverts the listen v2 regen (`4678f0e`, originally #750: force-end-turn, redaction, listen v1 diarize metadata) off `main`.

## Why
This work isn't ready to ship yet, and other priorities need to release from `main` without it. Since the regen is already on `main`, any release cut from `main` (including the pending release-please 7.7.0, #751) would publish it. Reverting removes it from the release line cleanly, without rewriting history.

## Preserved
The regen is not lost — it's parked on `gh/sdk-gen-2026-08-11` (branched from `main` at `4678f0e`) and can be re-applied / re-released when it's ready.

## Follow-ups
- The release-please PR #751 (release 7.7.0) should not be merged; once this lands, release-please will regenerate its release notes without the reverted work.
- To bring the regen back later, re-apply from `gh/sdk-gen-2026-08-11` (a revert-of-this-revert, or a fresh regen).


dg-coreylweathers (approved):
Approving. Verified this is a pure inverse of #750 — `git diff 7da415d 63e25d6` is empty, so the tree lands byte-identical to the `v7.6.0` tag. No collateral.

Checked independently rather than relying on CI:

- **No customer impact.** PyPI latest is 7.6.0; #750 bumped `pyproject` to 7.6.1 but never published. Rated against the last *released* SDK this is non-breaking — no semver event.
- **Gate passes on `63e25d6`:** pytest 331 passed / 1 skipped, mypy clean across 857 source files (run in a `python:3.12-slim` container, full wire suite via an external WireMock).
- **Release plumbing stays coherent.** The revert restores both halves of a matched pair — the `generic` `client_wrapper.py` entry in `release-please-config.json` *and* the `# x-release-please-version` annotations on the `User-Agent` / `X-Fern-SDK-Version` lines. No wire-telemetry version drift.
- **#758 restores what this drops:** reconnection plumbing, `importlib.metadata` version resolution, redaction, and all the compat-alias shims.

Two non-blocking notes on the description (code is fine, merge away):

**1. Scope is wider than the title.** #750 was a full generator cycle, not three spec features — it also carried stream-reconnection plumbing (`stream_reconnection_enabled` / `max_stream_reconnection_attempts`) and ~446 net lines across `src/deepgram/core/` (`http_sse/_api.py` alone +365). All of that reverts too, and all of it returns in #758. Worth a line in the body so the next reader of `main`'s history knows why `core/` moved.

**2. The "Preserved" paragraph points somewhere slightly wrong.** `gh/sdk-gen-2026-08-11` is stacked on *top* of this revert, and its tip is a fresh regen rather than a replay of #750 — grepping `src/deepgram` there gives 0 files for `force_end_turn` and 0 for `diarize_info` (redaction and the infra do survive). Nothing is orphaned, since `4678f0e` stays an ancestor of that branch, but "re-apply from the parked branch" would quietly hand someone a regen missing two of the three features. Suggest naming `4678f0e` as the recovery point for those two specifically.

**One thing to settle before the next release, not a blocker on this PR:** #751 currently announces "Listen v2 turn control & redaction" as shipped in 7.7.0, citing `4678f0e`. If it merges after this lands, we publish a changelog describing features that aren't in the tree. Suggest closing #751 now and letting #758 drive the next version, rather than letting release-please re-open it as a no-op 7.6.1 whose artifact is identical to 7.6.0. Your call on the release infra.

PR #768, opened 2026-08-18
feat(regen): listen v2 force-end-turn and listen v1 diarize metadata

GregHolmes:
Fern SDK regeneration for 2026-08-18 (fern-python-sdk `5.27.1`, CLI `5.95.1`).

**Status:** reconciled & verified. `mypy src/` clean (900 files) · `mypy tests/typecheck` clean · `ruff check src/deepgram` clean · **992 passed / 1 skipped** · coverage **94.69%** (≥ 90% gate). **No breaking changes** vs 7.7.0 — 0 removed public exports across every package, all renames aliased with identical values.

## New in this regen
- **Listen V1 diarization detail** — per-word `speaker` and `speaker_confidence` on the words item (pre-recorded only; `speaker_confidence` is not returned for streaming), plus `ListenV1ResponseMetadataDiarizeInfo` / `ListenV1ResultsMetadataDiarizeInfo` (`model_uuid`, `arch`) identifying which diarizer ran. Verified end-to-end against production: `diarize=true` returns `diarize_info` `{model_uuid, arch: "v1"}` and every word carries both fields; both are absent when diarization is off.
- **Listen V2 `eot_threshold` now accepts up to `1.0`** (previously documented as `0.5`–`0.9`). Boundary-verified against production: `0.49` and `1.01` are rejected with HTTP 400, `0.5` and `1.0` are accepted.
- **Listen V2 force-end-turn — requires server-side enablement, now live on staging.** `send_force_end_turn()` + `ListenV2ForceEndTurn` (a no-payload control message), plus a new `ListenV2TurnInfo.trigger` field (`model` | `manual` | `timeout`, open enum) identifying what ended a turn. Verified end-to-end on a deployment that has the feature; **still gated on production**, where sending the message returns `UNPARSABLE_CLIENT_MESSAGE` — *"The ForceEndTurn message is not enabled on this deployment."* — and closes the websocket. See [Force-end-turn and the end-of-turn controls](#force-end-turn-and-the-end-of-turn-controls).
- **Provider/agent-history type renames** — the generator dropped the `*ThinkProvider*` / `*SpeakProvider*` prefixes and renamed the agent conversation-history subtypes (see Backward compatibility). New canonical names: `AnthropicModel`, `CartesiaModelId`, `CartesiaVoice`, `DeepgramModel`, `GoogleModel`, `GoogleVersion`, `GroqReasoningMode`, `ConversationHistoryMessageRole`, `FunctionCallHistoryMessageFunctionCallsItem`.
- **`SpeakV2SpeedValue`** — a dedicated `float` type for the mid-stream `SpeakV2Configure.speed` field.

## Force-end-turn and the end-of-turn controls

**Updated 2026-08-20.** This supersedes the earlier finding that force-end-turn was rejected everywhere. The gate has since opened on staging, and the feature is now verified end-to-end through the SDK.

| Check (5s speech + 14s trailing silence unless noted) | Result |
|---|---|
| `send_force_end_turn()`, continuous speech | `EndOfTurn` with **`trigger="manual"`**, `end_of_turn_confidence=0.0085`, connection stays open and the turn index advances |
| `eot_threshold=0.7` | `EndOfTurn` with **`trigger="model"`** |
| `eot_threshold=1.0` | `EndOfTurn` with **`trigger="timeout"`** — *not* `"model"` |
| `eot_threshold=1.0` + `eot_timeout_ms=60000` | **no `EndOfTurn` at all** |
| same, then `send_force_end_turn()` | `EndOfTurn` with `trigger="manual"` |
| Production, any of the above | `UNPARSABLE_CLIENT_MESSAGE`, socket closed |

**The `eot_threshold=1.0` row corrects an earlier reading.** Counting events, `0.7` and `1.0` both produce exactly one `EndOfTurn`, which looks like `1.0` suppresses nothing. `trigger` shows that is a misattribution: at `1.0` Flux's confidence-based detection really is suppressed, and what closed the turn was `eot_timeout_ms` (default 5000), a separate mechanism. Nothing but `trigger` distinguishes those two cases — and `trigger` was not being emitted when this was last measured, which is why the distinction was invisible.

So the docstring's advice to "set to 1.0 to fully suppress natural end-of-turn detection and drive turn endings yourself" is **incomplete rather than wrong**: it needs `eot_timeout_ms` raised as well. The last two rows show the working combination. Worth reporting upstream as a spec-doc fix — the SDK type is generated and unfrozen, so patching the docstring here would mean freezing a file for a comment; better fixed at the source.

**No SDK change was needed** — `send_force_end_turn()` serialized correctly and `trigger` parsed all three values unmodified.

**Caveat:** one API key per environment, so this shows the gate is open *for that staging project*; whether enablement is per-project, per-deployment, or global still needs the feature-flag config.

## Backward compatibility (all preserved — non-breaking)
- **Renamed public types aliased** — old import paths kept, and the aliases resolve to the new types with byte-identical literal values: `AnthropicThinkProviderModel`→`AnthropicModel`, `CartesiaSpeakProviderModelId`→`CartesiaModelId`, `CartesiaSpeakProviderVoice`→`CartesiaVoice`, `DeepgramSpeakProviderModel`→`DeepgramModel`, `GoogleThinkProviderModel`→`GoogleModel`, `GoogleThinkProviderVersion`→`GoogleVersion`, `GroqThinkProviderReasoningMode`→`GroqReasoningMode`.
- **Agent-history names shimmed** — the removed `AgentV1HistoryContentRole`, `AgentV1HistoryFunctionCallsFunctionCallsItem` (+`Params`), and the double-`Context` `...ContentRole` / `...FunctionCallsFunctionCallsItem` (+`Params`) are recreated as hand-written aliases to the new canonical types, so every old import path still resolves.
- **Repointed shims** — the single-`Context` `...MessagesItemContentRole` / `...FunctionCallsFunctionCallsItem` shims whose underlying generated modules were deleted now re-export the new canonical types.
- `send_force_end_turn()` takes an optional message (no-payload control), matching its siblings; listen/v2 `send_configure` still accepts a raw dict alongside the typed model; callable `AgentV1SettingsAgent(...)`, legacy `messages=` / `.messages`, and `language_hint=` all continue to work.

## SpeakV2Speed retype — patched (spec bug, raised with Fern)
The generator retyped `SpeakV2Speed` (the `speak.v2.connect(speed=...)` param) from `float` to `Union[Literal["0.85"…"1.15"], Any]` — a string-literal enum. That contradicts the actual API contract and is inconsistent with the new `SpeakV2SpeedValue = float` on the Configure message. Verified against production (`wss://api.deepgram.com`, `flux-alexis-en`):
- `speed: 1.05` (number) → `ConfigureSuccess {applied.speed: 1.05}`
- `speed: 2.0` → `ConfigureFailure SPEED_OUT_OF_RANGE` (value echoed back as the number `2.0`)
- `speed: "1.05"` (string) → `Error MESSAGE-0000 "The message could not be parsed."`

So the string form the generated type implies is rejected at the wire. Patched `SpeakV2Speed` back to `float` (frozen in `.fernignore`, guarded by a regression test). Raised upstream with Fern; unfreeze and drop the patch once the spec/generator emits a number.

## Maintenance
**Patches re-applied** (generator still needs them): socket-client broad `except` + optional no-payload control params; agent `_sanitize_numeric_types`; `query_encoder` bool coercion; `ApiError` / `ParsingError` credential redaction; agent-settings and `language_hint`→`language_hints` back-compat validators; `AgentV1UpdateListenListen` legacy-provider coercion; the legacy alias re-exports merged into 10 `__init__.py`; both frozen wire tests.

**New frozen files this cycle:** 14 back-compat alias modules for the renames, plus `types/speak_v2speed.py` (the `float` patch). `.fernignore` and the AGENTS.md freeze ledger updated to match.

**Added test coverage:** `send_force_end_turn` (sync + async), listen/v2 `send_configure` typed-model + async paths, the provider/agent-history rename aliases, and the `SpeakV2Speed`-is-`float` regression guard. Plus live coverage in `tests/manual/listen/v2/force_end_turn/` (sync + async, since the two socket clients have separate implementations) and `examples/16-transcription-force-end-turn.py`. All three report the gate and exit cleanly where the feature is not enabled, rather than dying on the connection teardown the rejection causes.

## Verification
- Unit: `mypy` + `ruff` clean, 992 passed / 1 skipped, coverage 94.69%.
- Integration: 37/37 `tests/manual` scripts pass against the live API, plus the two new `force_end_turn` scripts — all four checks pass against a deployment with the feature, and both report SKIP against production.
- Examples: 21/22 non-management examples pass; the only failure is `27-transcription-live-sagemaker.py`, which needs the optional `deepgram_sagemaker` package (not a regression). New example 16 verified against both a feature-enabled deployment and production.

### Version bump
The branch carries a patch bump in `pyproject.toml`, but this is a `feat`. That's expected — the generator patch-bumps on every regen and release-please recomputes the correct minor at release time from the conventional-commit type. This only holds because the squash-merge message is `feat(regen): …`, so the PR title matters here.


dg-coreylweathers (approved):
Reviewed the regen diff, re-applied patches, freeze-ledger updates, and the reference.md send_force_end_turn addition — CI green on a951216, no breaking changes vs 7.7.0. LGTM.

GregHolmes:
@dg-coreylweathers Follow-up on the Flux speed rollout: production verification on 2026-09-10 accepted all four boundary values: `0.5`, `0.85`, `1.15`, and `1.5`.

- JS SDK: batch REST synthesis returned non-empty audio for each value; Speak V2 WebSocket `Configure` returned `ConfigureSuccess` and echoed the exact requested `applied.speed` for each.
- Python SDK: the same REST and WebSocket matrix passed.
- Java SDK: the four-value Speak V2 WebSocket `Configure` matrix passed with exact echoed values; its existing live streaming-audio integration also passed at `1.05`. I did not run a separate Java REST boundary matrix.

No `SPEED_OUT_OF_RANGE` responses occurred at either new boundary. This resolves the prior fleet-rollout acceptance concern. A separate repeated batch-audio duration probe showed substantial variation at `0.5`, so I would not use these results alone to claim strict rate-fidelity at the slow endpoint.

PR #769, opened 2026-08-21
fix(listen/v2): restore subscript access on typed responses

GregHolmes:
## Summary
- restore read-only `response["field"]` access only on the seven Listen V2 response and nested-response model classes affected by the 7.7.0 retype
- preserve canonical attribute access, nested word subscripting, unknown fields, wire aliases, and `KeyError` for omitted fields
- leave all unrelated generated models unchanged
- freeze the hand-written compatibility base and seven generated response classes for future regens

## Why
SDK 7.7.0 removed the `typing.Any` member from `V2SocketClientResponse`, correctly restoring typed Listen V2 responses. Through 7.6, that member caused every Listen V2 response to be returned as a raw dictionary, so callers using `response["field"]` broke when upgrading.

This surfaced while reviewing [deepgram-docs#1130](https://github.com/deepgram/deepgram-docs/pull/1130). The docs should continue teaching typed attribute access; this narrow shim protects the previously working subscript syntax.

Both styles now work:

```python
message.words[0].confidence
message["words"][0]["confidence"]
message.words[0]["confidence"]
```

This PR intentionally does not turn Pydantic models into full dictionaries or `Mapping` objects. It restores only the subscript access that triggered the customer-facing regression. Models remain immutable and attribute access remains canonical.

## Verification
- `pytest`: 985 passed, 1 skipped
- `mypy src/`: clean
- `mypy tests/typecheck`: clean
- `ruff check src/deepgram tests/custom/test_model_dict_compat.py`: clean
- targeted compatibility tests pass with Pydantic 1.10.18 and Pydantic 2.x
- sync and async receive paths, all Listen V2 response variants, nested response objects, unknown fields, omitted declared fields, and unrelated-model isolation are covered
- standalone probe passes on this branch and fails on `main` with `TypeError: ListenV2TurnInfo object is not subscriptable`


dg-edcharbeneau (dismissed):
Greg's fix addresses the removes the TypeError. **#1130 should now be closed, not merged**

dg-edcharbeneau (approved):

deepgram-kiley (approved):
Approving based on @dg-edcharbeneau's previous review.

Issue #734, opened 2026-06-25
`DeepgramClient()` reads `DEEPGRAM_API_KEY` at import time, not at instantiation


samgutentag:
# Summary

`api_key` defaults to `os.getenv("DEEPGRAM_API_KEY")` directly in the `__init__` signature, so Python evaluates it once, when the module is imported, not when the client is constructed. The idiomatic pattern (imports at the top of a file, `load_dotenv()` below them) therefore fails: the default is captured as `None` before the key is set, and `DeepgramClient()` raises even though `DEEPGRAM_API_KEY` is present by the time you instantiate.

# Environment

- deepgram-sdk **7.3.1**
- Python 3.14.6
- macOS (Darwin 24.6.0)

# Root cause

`deepgram/base_client.py` (Fern-generated), `BaseClient.__init__`:

```python
api_key: typing.Optional[str] = os.getenv("DEEPGRAM_API_KEY"),   # evaluated at import
...
if api_key is None:
    raise ApiError(body="The client must be instantiated be either passing in api_key or setting DEEPGRAM_API_KEY")
```

`os.getenv(...)` as a default argument runs once, at function-definition (import) time, not per call.

# Reproduction

```python
import os
os.environ.pop("DEEPGRAM_API_KEY", None)         # not set in the shell at process start
from deepgram import DeepgramClient               # default captured as None HERE
os.environ["DEEPGRAM_API_KEY"] = "dummy-key"      # set afterward (what load_dotenv() does)
DeepgramClient()                                  # raises, despite the var being set
```

Output on 7.3.1:

```
set at instantiation: True
DeepgramClient() -> ApiError: ... The client must be instantiated be either passing in api_key or setting DEEPGRAM_API_KEY
DeepgramClient(api_key=...) -> constructed OK
```

# Impact

The documented zero-config pattern (`DeepgramClient()` reading `DEEPGRAM_API_KEY`) only works if the variable is in the OS environment *before* the `deepgram` import runs. With `python-dotenv` (very common), `load_dotenv()` runs after the import, so the key is invisible. The error message ("set DEEPGRAM_API_KEY") is misleading because the variable is already set, which sends you to debug the wrong thing.

### Suggested fix

Default the parameter to `None` and resolve the env var inside `__init__` so it's read at call time:

```python
def __init__(self, *, api_key: typing.Optional[str] = None, ...):
    if api_key is None:
        api_key = os.getenv("DEEPGRAM_API_KEY")
    if api_key is None:
[...]

PR #778, opened 2026-09-01
fix: preserve explicit API key semantics

GregHolmes:
## Summary

- resolve `DEEPGRAM_API_KEY` at construction only when the `api_key` argument is omitted
- preserve the existing behavior where explicit `api_key=None` rejects authentication instead of silently using an ambient process credential
- isolate the import-time environment regression in a subprocess that starts without `DEEPGRAM_API_KEY`
- add sync and async coverage for explicit-`None` behavior

Follow-up to #767 and #734.

## Why

#767 correctly fixed delayed environment loading, but its `kwargs.get("api_key") is None` check treated an omitted argument and explicit `None` as equivalent. In tests, multi-tenant services, or applications intentionally suppressing ambient authentication, explicit `None` must not activate a process-wide credential.

## Verification

- full Docker-backed suite: `1020 passed, 1 skipped`
- `mypy src`: clean across 901 source files
- Ruff: clean for both changed files
- seven-scenario sync/async credential probe: all passed
- regression tests proven non-vacuous against both the pre-#767 baseline and the original #767 implementation
- `git diff --check`: clean

### DIFF
--- a/src/deepgram/client.py
+++ b/src/deepgram/client.py
@@ -115,7 +115,7 @@ class DeepgramClient(BaseClient):
             # Set a placeholder api_key if none provided (base client requires it)
             if kwargs.get("api_key") is None:
                 kwargs["api_key"] = "token"
-        elif kwargs.get("api_key") is None:
+        elif "api_key" not in kwargs:
             # The generated base client takes os.getenv("DEEPGRAM_API_KEY") as a default
             # argument, so it is read once at import. Re-read it here so a key set after
             # import (load_dotenv below the imports) is still picked up.
@@ -199,7 +199,7 @@ class AsyncDeepgramClient(AsyncBaseClient):
             # Set a placeholder api_key if none provided (base client requires it)
             if kwargs.get("api_key") is None:
                 kwargs["api_key"] = "token"
-        elif kwargs.get("api_key") is None:
+        elif "api_key" not in kwargs:
             # The generated base client takes os.getenv("DEEPGRAM_API_KEY") as a default
             # argument, so it is read once at import. Re-read it here so a key set after
             # import (load_dotenv below the imports) is still picked up.
--- a/tests/custom/test_api_key_env_resolution.py
+++ b/tests/custom/test_api_key_env_resolution.py
@@ -8,6 +8,10 @@ client's re-read so the common ``load_dotenv()``-below-the-imports layout keeps
 working (issue #734).
 """
 
+import os
+import subprocess
+import sys
+import textwrap
 import typing
 
 import pytest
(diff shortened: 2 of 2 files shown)


dg-coreylweathers (approved):
**What this PR does**

PR #767 made the SDK re-read `DEEPGRAM_API_KEY` when the client is constructed, but its check treated an omitted `api_key` argument and an explicit `api_key=None` as the same thing, so passing `None` on purpose silently picked up whatever key was in the process environment. This PR changes the check (`"api_key" not in kwargs`) in both the sync and async clients so explicit `None` is rejected with an `ApiError` at construction — which is exactly what the released 7.8.0 does today.

**What I checked**

- **Does explicit `None` actually raise?** Yes — it now passes through to the generated base client, which raises `ApiError` when the key is `None`, and that client's parameters are keyword-only, so the membership check cannot be bypassed positionally.
- **Did the regression ever reach a release?** No — #767 merged after 7.8.0 shipped, so no released version ever had the explicit-`None`-falls-back-to-env behavior; #767 and this fix will ship together.
- **Are the new tests real?** Yes — the rewritten load_dotenv-ordering test spawns a subprocess with the key stripped from the environment, which is the only way to genuinely reproduce "imported without a key" (the old in-process test could not). The new explicit-`None` test fails on current `main`, so it proves the fix rather than restating it.
- Everything else passed: CI green across Python 3.10–3.13, the `access_token` path is unchanged and still covered, and no README/docs/example anywhere describes `api_key=None` as an environment-fallback idiom.

No changes requested.

dg-coreylweathers (approved):
**What this PR does**

PR #767 made the SDK re-read `DEEPGRAM_API_KEY` when the client is constructed, but its check treated an omitted `api_key` argument and an explicit `api_key=None` as the same thing, so passing `None` on purpose silently picked up whatever key was in the process environment. This PR changes the check (`"api_key" not in kwargs`) in both the sync and async clients so explicit `None` is rejected with an `ApiError` at construction — which is exactly what the released 7.8.0 does today.

**What I checked**

- **Does explicit `None` actually raise?** Yes — it now passes through to the generated base client, which raises `ApiError` when the key is `None`, and that client's parameters are keyword-only, so the membership check cannot be bypassed positionally.
- **Did the regression ever reach a release?** No — #767 merged after 7.8.0 shipped, so no released version ever had the explicit-`None`-falls-back-to-env behavior; #767 and this fix will ship together.
- **Are the new tests real?** Yes — the rewritten load_dotenv-ordering test spawns a subprocess with the key stripped from the environment, which is the only way to genuinely reproduce "imported without a key" (the old in-process test could not). The new explicit-`None` test fails on current `main`, so it proves the fix rather than restating it.
- Everything else passed: CI green across Python 3.10–3.13, the `access_token` path is unchanged and still covered, and no README/docs/example anywhere describes `api_key=None` as an environment-fallback idiom.

No changes requested.

PR #780, opened 2026-09-02
docs(examples): add microphone live transcription example

### LINKED ISSUE #779: docs(examples): add microphone live transcription example
## Goal

Provide a maintained, copy-pasteable Python example that streams microphone audio to Listen V1 without adding microphone capture or platform audio dependencies to the core SDK.

## Scope

- Add `examples/18-transcription-live-microphone.py` using the optional `sounddevice` dependency.
- Explain installation, device selection, and audio format settings in the script.
- Read `DEEPGRAM_API_KEY` from the environment and use `DeepgramClient` plus the public Listen V1 socket API.
- Send linear16 microphone bytes at the matching sample rate and channel count.
- Handle transcript, error, close, Ctrl-C, and CloseStream paths cleanly.
- Add the example to `examples/README.md`.

## Non-goals

- No microphone code, optional extra, or audio-device dependency in `deepgram-sdk`.
- No cross-platform device abstraction beyond what sounddevice provides.
- No automatic reconnection or application-specific audio pipeline.

## Acceptance

- The script imports only after the user installs `sounddevice`.
- It runs against a real microphone with documented setup.
- It uses public SDK APIs only and does not expose credentials.

GregHolmes:
Closes #779

## Summary

- add a maintained Listen V1 microphone transcription example using the optional `sounddevice` dependency
- support device listing, device selection by name or numeric ID, sample rate, channel count, block duration, and language configuration
- decouple the PortAudio callback from WebSocket I/O with a bounded queue, reporting dropped chunks when the sender cannot keep up
- handle final transcript events, SDK errors, Ctrl-C, Finalize, and CloseStream cleanly
- list the new example in `examples/README.md`

## Scope

This is an example only. It adds no core microphone capture abstraction, optional package extra, or platform-audio dependency to `deepgram-sdk`; microphone drivers, permissions, and backend installation remain owned by `sounddevice` and the host platform.

## Validation

- live macOS microphone run with device ID 2: received final Deepgram transcripts and returned cleanly after Ctrl-C
- verified the missing `sounddevice` path prints the install command without a traceback
- verified `--list-devices` with an isolated fake backend
- Python compilation, Ruff, and mypy pass
- existing SDK suite: `1019 passed, 1 skipped`
- `git diff --check` passes

### DIFF
--- a/examples/18-transcription-live-microphone.py
+++ b/examples/18-transcription-live-microphone.py
@@ -1,0 +1,197 @@
+"""Live microphone transcription with Listen V1.
+
+Install the optional capture dependency first:
+
+    pip install sounddevice
+
+On Linux, the system PortAudio library may also be required. List available
+input devices before recording with:
+
+    python examples/18-transcription-live-microphone.py --list-devices
+
+Then set DEEPGRAM_API_KEY and run:
+
+    python examples/18-transcription-live-microphone.py
+
+This example sends 16-bit PCM (linear16) to Deepgram. The sample rate and
+channel count in the WebSocket connection always match the microphone stream.
+"""
+
+import argparse
+import os
+import queue
+import sys
+import threading
+from typing import Any, Optional
+
+from dotenv import load_dotenv
+
+from deepgram import DeepgramClient
+from deepgram.core.events import EventType
+from deepgram.listen.v1.types import ListenV1Results
+
+load_dotenv()
+
+MAX_QUEUED_CHUNKS = 20
+
+
+def load_sounddevice() -> Any:
+    try:
+        import sounddevice  # type: ignore[import-not-found]
+    except ImportError as exc:
+        raise RuntimeError(
+            "Microphone capture requires sounddevice. Install it with: pip install sounddevice"
+        ) from exc
+    return sounddevice
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description="Transcribe microphone audio with Deepgram Listen V1")
+    parser.add_argument("--list-devices", action="store_true", help="list microphone devices and exit")
+    parser.add_argument("--device", help="input device name or numeric ID from --list-devices")
+    parser.add_argument("--sample-rate", type=int, default=16000, help="input sample rate in Hz (default: 16000)")
+    parser.add_argument("--channels", type=int, default=1, help="input channel count (default: 1)")
+    parser.add_argument(
+        "--block-duration",
+        type=float,
+        default=0.1,
+        help="audio callback block duration in seconds (default: 0.1)",
+    )
+    parser.add_argument("--language", default="en", help="recognition language (default: en)")
+    return parser.parse_args()
+
+
+def main() -> int:
+    args = parse_args()
+    try:
+        sounddevice = load_sounddevice()
+    except RuntimeError as exc:
+        print(exc, file=sys.stderr)
+        return 2
+
+    if args.list_devices:
+        print(sounddevice.query_devices())
+        return 0
+
+    if not os.getenv("DEEPGRAM_API_KEY"):
+        print("Set DEEPGRAM_API_KEY before recording.", file=sys.stderr)
+        return 2
+    if args.sample_rate <= 0 or args.channels <= 0 or args.block_duration <= 0:
+        print("sample-rate, channels, and block-duration must be positive.", file=sys.stderr)
+        return 2
+
+    blocksize = int(args.sample_rate * args.block_duration)
+    if blocksize == 0:
+        print("block-duration is too small for the selected sample rate.", file=sys.stderr)
+        return 2
+
+    device: Optional[object] = int(args.device) if args.device and args.device.isdigit() else args.device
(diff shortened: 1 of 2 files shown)


dg-coreylweathers (changes requested):
**What this PR does**

Adds example 18: live microphone transcription through Listen V1 using the optional `sounddevice` package, plus its line in `examples/README.md`. The microphone callback hands audio to a bounded 20-slot queue that a separate thread drains, so a slow network can never block the audio driver; dropped chunks are counted and reported, and Ctrl-C sends Finalize + CloseStream before exiting.

**What I checked**

- **Does every SDK call it makes exist?** Yes — all seven `connect()` parameters and every method (`send_media`, `send_finalize`, `send_close_stream`, `start_listening`), event, and response field it uses match the Listen V1 surface by name.
- **Is the threading sound?** Yes — I traced callback → queue → sender thread → error propagation → shutdown; sender failures and SDK error events both stop the run and surface as a non-zero exit, and both worker threads are daemons so exit can't hang.
- **Can the API key leak?** No — the key comes from the environment only and is never printed.
- **Not verified here:** the live macOS microphone run is author-attested; a real mic can't be exercised headlessly.

**One fix before merge**

When the connection fails (a bad key is the most common case), the developer sees only `Microphone transcription failed: ApiError` — the status code and message are discarded. This SDK's `ApiError` already masks the `Authorization` header at construction, so printing the exception is safe and makes the failure actionable. In `examples/18-transcription-live-microphone.py`, line ~183:

```python
print(f"Microphone transcription failed: {type(exc).__name__}: {exc}", file=sys.stderr)
```

GregHolmes:
Thank you @dg-coreylweathers  could you please re-review this?

dg-coreylweathers (changes requested):
**What this PR does**

Adds example 18: live microphone transcription through Listen V1 using the optional `sounddevice` package, plus its line in `examples/README.md`. Closes #779. No SDK code changes.

**Re-review of the earlier fix**

The requested change is in (f600496): connection failures now print the exception type and message. I ran the script with a deliberately bad key and got `Microphone transcription failed: InvalidStatus: server rejected WebSocket connection: HTTP 401`, exit code 1, with the key absent from all output. One note on my earlier wording: the SDK surfaces a rejected handshake as the websockets library's `InvalidStatus`, not `ApiError`, because the generated compat shim binds the wrong exception class. Your fix handles either, so nothing to change there.

The widened `type: ignore[import-not-found,import-untyped]` is correct: mypy reports the first code when sounddevice is absent and the second when it is installed without stubs. Both environments pass under mypy 1.13.

**One fix before merge**

On Debian or Ubuntu with sounddevice installed but the PortAudio system library missing, `import sounddevice` raises `OSError`, not `ImportError`. The script's friendly install hint only catches `ImportError`, so the developer gets a full traceback and exit 1. I reproduced this in a `python:3.12` container. Fix is to catch both and name the apt package. In `examples/18-transcription-live-microphone.py`, line ~41:

```python
    except (ImportError, OSError) as exc:
        raise RuntimeError(
            "Microphone capture requires sounddevice and the PortAudio library. "
            "Install them with: pip install sounddevice "
            "(on Debian/Ubuntu also: sudo apt install libportaudio2)"
        ) from exc
```

**What I checked**

- Every `connect()` argument, socket method, event type, and response field the script uses exists on the Listen V1 surface.
- Every `RawInputStream` argument exists in sounddevice 0.5.6.
- Sample rate and channel count flow from the CLI args into both the mic stream and the websocket, never hardcoded.
- ruff, mypy, py_compile, `git diff --check`, and CI on 3.10-3.13 all pass.
- Not verified here: the live macOS microphone run (author-attested; no mic headlessly).

**Nits, optional**

- `on_close` (line 117) sets the stop flag silently, so a clean server close exits 0 with no output after "Recording..."; example 13 prints "Connection closed". Suggest printing the same.
- README entry: "Live microphone transcription (Listen V1) with the optional sounddevice dependency", matching how entries 13 and 14 name the API version.

dg-coreylweathers (approved):
**What this PR does**

Adds example 18: live microphone transcription through Listen V1 using the optional `sounddevice` package, plus its line in `examples/README.md`. Closes #779. No SDK code changes.

**Re-review of the requested fix**

Commit 3157d39 applies the PortAudio change as suggested: `load_sounddevice()` now catches `OSError` alongside `ImportError` and names `libportaudio2` in the hint. I reproduced the Debian case in a `python:3.12` container with `sounddevice` installed and no PortAudio library on the system: the script prints the one-line install hint, exits 2, no traceback. Same result for `--list-devices` in that environment and for the no-`sounddevice` environment. Only those four lines changed since the last review.

**What I checked**

- ruff check and ruff format pass; mypy 1.13 passes with and without `sounddevice` installed; py_compile passes.
- CI at 3157d39: compile and test on 3.10-3.13, Title Check, all green.
- Not verified here: the live macOS microphone run (author-attested; no mic headlessly).

The two earlier nits (printing "Connection closed" in `on_close`, and naming Listen V1 in the README entry) were not taken and remain optional. Approving.

PR #781, opened 2026-09-03
docs(examples): add Voice Agent session recorder

### LINKED ISSUE #775: Example: Record a Voice Agent session as JSON
## Summary

Add a standalone Voice Agent example that demonstrates application-owned session recording. The example collects selected server events during one session and serializes a JSON record when the session finishes.

## Why an example

Session recording, storage, consent, retention, and redaction policies belong to each application. The SDK should continue to expose the Voice Agent event stream without retaining customer conversation data or adding a persistence abstraction.

## Example behavior

- Collect received conversation text, function-call requests and responses, and latency-report events when present.
- Build a JSON-serializable session record after the connection closes.
- Record only data delivered by the server; do not invent speaker attribution, audio segment references, or per-component latency fields.
- Do not capture or store audio.
- Include a clear warning that transcripts and tool arguments/results can contain sensitive data; applications must apply their own consent, redaction, retention, and storage policy.

## Acceptance criteria

- [ ] Runs as a standalone Voice Agent example using the current SDK event API.
- [ ] Demonstrates an application-owned recorder and JSON serialization.
- [ ] Does not add runtime dependencies or change generated SDK code.
- [ ] Documents the data-handling caveat in the example.

GregHolmes:
## Summary
- add an application-owned Voice Agent event recorder example
- serialize received transcript, function-call, and latency events as JSON
- document that transcripts and tool data need application consent, redaction, retention, and storage controls

## Validation
- `poetry run ruff format --check examples/31-voice-agent-session-recording.py`
- `poetry run ruff check examples/31-voice-agent-session-recording.py`
- `poetry run mypy --ignore-missing-imports examples/31-voice-agent-session-recording.py`
- `poetry run python -m py_compile examples/31-voice-agent-session-recording.py`
- `poetry run mypy src/`
- `poetry run mypy tests/typecheck`
- `poetry run pytest -rP --cov=deepgram --cov-branch --cov-report=xml --cov-report=term-missing .`

Closes #775

### DIFF
--- a/examples/31-voice-agent-session-recording.py
+++ b/examples/31-voice-agent-session-recording.py
@@ -1,0 +1,146 @@
+"""
+Example: Record a Voice Agent session as JSON
+
+This stores received transcripts, function-call events, and latency reports in
+voice-agent-session.json after the session ends. Those fields can contain
+sensitive data. Apply your application's consent, redaction, retention, and
+storage policies before using this pattern in production.
+
+The recorder does not capture audio or manufacture fields that the server did
+not send, such as audio references or unavailable latency measurements.
+"""
+
+import json
+import threading
+import time
+from dataclasses import dataclass, field
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Any
+
+from dotenv import load_dotenv
+
+from deepgram import DeepgramClient
+from deepgram.agent.v1.types import (
+    AgentV1InjectUserMessage,
+    AgentV1Settings,
+    AgentV1SettingsAgent,
+    AgentV1SettingsAgentListen,
+    AgentV1SettingsAgentListenProvider_V1,
+    AgentV1SettingsAudio,
+    AgentV1SettingsAudioInput,
+)
+from deepgram.core.events import EventType
+from deepgram.types.speak_settings_v1 import SpeakSettingsV1
+from deepgram.types.speak_settings_v1provider import SpeakSettingsV1Provider_Deepgram
+from deepgram.types.think_settings_v1 import ThinkSettingsV1
+from deepgram.types.think_settings_v1provider import ThinkSettingsV1Provider_OpenAi
+
+load_dotenv()
+
+OUTPUT_PATH = Path("voice-agent-session.json")
+USER_MESSAGE = "What does the first all-female spacewalk symbolize?"
+
+
+def _serialize_event(message: Any) -> dict[str, Any]:
+    if hasattr(message, "model_dump"):
+        return message.model_dump(mode="json")
+    return json.loads(message.json())
+
+
+@dataclass
+class SessionRecorder:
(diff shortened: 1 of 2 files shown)


dg-coreylweathers (approved):
**What this PR does**

Adds example 31: a standalone Voice Agent script that injects one text question, collects the transcript, function-call, and latency events the server sends, and writes them to one JSON file when the session ends. Closes #775. No SDK code changes.

**What I checked**

- **Does it run end to end?** Yes. In a `python:3.12` container with a valid key it finished in about 3 seconds and wrote a record with the user question, the assistant reply, and four latency reports (`ttt_token_latency`, `ttt_text_latency`, `tts_latency`, `total_latency`).
- **Is the temp-file claim true?** Yes. The fallback temp file is created mode 0600. A file passed via `--output` gets the default umask (0644 in my run).
- **Can the API key leak?** No. The key is absent from stdout, stderr, and the record; a fake-key run exposes nothing either.
- **Do the event names and imports match the SDK?** Yes. All six imported types exist, and `ConversationText`, `FunctionCallRequest`, `FunctionCallResponse`, and `LatencyReport` are all in the server response union.
- ruff format, ruff check, mypy, py_compile, and CI on 3.10-3.13 pass. No new runtime dependency.

**Should-fix (non-blocking, can follow in a later commit)**

1. A server `Error` event is silently ignored. `on_message` (line ~130) handles only `SettingsApplied` and `ConversationText`, so if the think provider fails after the question is injected, the script waits the full 30 seconds and exits with "Timed out waiting for the agent to respond" while the real cause is lost. Example 30 prints `Error` events. Suggest adding, after the `ConversationText` branch:

```python
            elif getattr(message, "type", None) == "Error":
                print(f"Agent error: {message.code} - {message.description}")
                assistant_response_event.set()
```

and raising with the server's message after the wait instead of a timeout.

2. A failed or timed-out session writes no record. Both `TimeoutError` raises (lines ~151 and ~155) exit the `with` block before `recorder.finish()` and the write, so the events already collected are lost and `ended_at` stays null. The docstring says "after a successful session," so this is documented, but the failed session is the one a developer most wants to inspect. Suggest a `try/finally` around the `with` block that always writes the record and lets the exception propagate.

**Nits, optional**

- Docstring (lines 1-13): the settings define no functions, so `function_calls` is always `[]` in a real run. One sentence saying so would set expectations.
- Line ~113: a bad or missing key produces a raw traceback ending in `InvalidStatus: HTTP 401`; examples 18 and 30 print one line. Wrapping `main()` in `except Exception as exc: print(f"Session failed: {type(exc).__name__}: {exc}", file=sys.stderr)` would match them. The key is not in the traceback.
- Line ~57: `--output` files get default permissions while the temp fallback is 0600 and the docstring stresses sensitive contents. Either open with `0o600` or say in the docstring that `--output` permissions are the caller's.

